PR #2, opened 2020-06-19
Featuring CameraX api for camera handling

aquatiko:
Adding the boilerplate code inspired by [here](https://github.com/googlesamples/mlkit/tree/master/android/vision-quickstart). Will add changes to suit my application in following commits.

## Why the shift to CameraX?
- CameraX was easier to use.
- It's built on top of Camera2 api which gives more fine control over camera than Camera1 api.
- Support for previous APIs are dropping in the community.


vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -164,15 +169,22 @@ public void processImageProxy(ImageProxy image, GraphicOverlay graphicOverlay) {
...
-                /* originalCameraImage= */ bitmap,
-                /* shouldShowFps= */ true)
-                // When the image is from CameraX analysis use case, must call image.close() on received
-                // images when finished using them. Otherwise, new images may not be received or the camera
-                // may stall.
-                .addOnCompleteListener(results -> image.close());
+        if(bufferImage.size()>depth)
+            bufferImage.clear();        // O(n) operation
+
+        bufferImage.add(new Pair(image.getImage(), System.currentTimeMillis()));
```
aquatiko: time difference between consecutive buffer addition takes around 20-40 ms.

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -164,15 +169,22 @@ public void processImageProxy(ImageProxy image, GraphicOverlay graphicOverlay) {
...
-                InputImage.fromMediaImage(image.getImage(), image.getImageInfo().getRotationDegrees()),
-                graphicOverlay,
-                /* originalCameraImage= */ bitmap,
-                /* shouldShowFps= */ true)
-                // When the image is from CameraX analysis use case, must call image.close() on received
-                // images when finished using them. Otherwise, new images may not be received or the camera
-                // may stall.
-                .addOnCompleteListener(results -> image.close());
+        if(bufferImage.size()>depth)
+            bufferImage.clear();        // O(n) operation
```
aquatiko: flushing operation dosen't seem to impart significant time lag.

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -80,6 +83,8 @@
     private ByteBuffer processingImage;
     @GuardedBy("this")
     private FrameMetadata processingMetaData;
+    private ArrayList<Pair<Image, Long>> bufferImage = new ArrayList<Pair<Image, Long>>();
```
mwarton: Can you name this something that makes it clear that it is a collection, eg bufferImageList?  This makes it much easier to read the code later on, I keep thinking it is a single Image.

mwarton: OK for experimentation, but if this becomes part of the app, please define an inner class rather than a Pair so you can name the fields

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
mwarton: isn't this just the time between new frames from the camera?  You aren't doing anything with the images here.

aquatiko: @mwarton I expected it that way and it should have been around ~16 ms for a 60fps camera. I wanted to test the processing delay between two consecutive buffer addition. Contrary to my expectation it was around 20-40 ms. Could be due to memory copy operations.

mwarton:
This is in the preview lifecycle method though right?  It would be totally
expected to drop frames for the preview since the human eye can only see
~25 fps anyway

aquatiko:
###  Buffer test experiment
It was not possible to add `Image` or `ImageProxy` to buffer and use it later (as we are closing them moments after reading them).
I converted it to `Bitmap` and stored in buffer. 
To test if I can use the buffered frame, I tasked `GraphicOverlay` (preview) to display `0th` frame only, while the new frames are still coming and getting buffered.
When the `depth` of buffer is reached, the buffer is cleared out and a new `0th` frame is set to `GraphicOverlay`.

To test the speed of the copy operation on frames, set `depth=2` and you can see the responsiveness in the `GraphicOverlay` updates. I didn't see aparent preview update lags.

aquatiko:
Note: the function `tobitmap(...)` need to be fixed. 
It's borrowed from stackOverflow and shows an incorrect rotation in preview.
Also, while closing frames, we would need to get all the required frame info(eg. height, width, rotationdegree) from `ImageProxy` and store it in a class as we can't access them later.(after closing)

**Update:**
Fixed above problems in refactoring commit.

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -153,7 +158,30 @@ private void processImage(
...
+
+        byte[] nv21 = new byte[ySize + uSize + vSize];
+        //U and V are swapped
+        yBuffer.get(nv21, 0, ySize);
+        vBuffer.get(nv21, ySize, vSize);
+        uBuffer.get(nv21, ySize + vSize, uSize);
+
+        YuvImage yuvImage = new YuvImage(nv21, ImageFormat.NV21, image.getWidth(), image.getHeight(), null);
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        yuvImage.compressToJpeg(new Rect(0, 0, yuvImage.getWidth(), yuvImage.getHeight()), 75, out);
```
mwarton: Why are you converting to jpeg and then to bitmap?  You are doing an uneccesary encode/decode step and probably losing detail in the process.  Can't you just buffer (and use) the yuvimages directly?

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -163,18 +191,28 @@ public void processImageProxy(ImageProxy image, GraphicOverlay graphicOverlay) {
...
-        if(bufferImage.size()>depth)
-            bufferImage.clear();        // O(n) operation
-
-        bufferImage.add(new Pair(image.getImage(), System.currentTimeMillis()));
+        Log.d("Rokus logs", "Current buffer size:" + bufferImageList.size());
+        graphicOverlay.clear();
+        if (bitmap != null) {
+            try {
+                graphicOverlay.add(new CameraImageGraphic(graphicOverlay, bitmap));
+            } catch (Exception e) {
```
mwarton, line 198: Can you use a more specific exception?  what commonly goes wrong here?

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -88,8 +89,19 @@
...
+    {
+        Bitmap image;
+        ImageInfo info;
+        bufferImage(Bitmap image, ImageInfo info)
+        {
+            this.image = image;
+            this.info = info;
+        }
+    };
+    private ArrayList<bufferImage> bufferImageList = new ArrayList<>();
```
mwarton, line 103: Initialize the array list with the correct capacity, since you know it in advance.

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -88,8 +89,19 @@
     private ByteBuffer processingImage;
     @GuardedBy("this")
     private FrameMetadata processingMetaData;
-    private ArrayList<Pair<Bitmap, Long>> bufferImageList = new ArrayList<>();
-    private Integer depth = 30;
+
+    class bufferImage
```
mwarton, line 93: private class?

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -193,13 +182,12 @@ public void processImageProxy(ImageProxy image, GraphicOverlay graphicOverlay) {
         }
         if(bufferImageList.size()>depth)
             bufferImageList.clear();        // O(n) operation
-
-        bufferImageList.add(new Pair(toBitmap(image.getImage()), System.currentTimeMillis()));
+        bufferImageList.add(new bufferImage(BitmapUtils.getBitmap(image), image.getImageInfo()));
```
mwarton, line 185: Did you forget to add the file for BitmapUtils?

Also Utils is usually a bad name for things, try to name it for what it actually does, otherwise you end up with giant classes containing lots of unrelated things.  In this case BitmapConverter or something is probably more appropriate.

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
```diff
@@ -88,8 +89,19 @@
     private ByteBuffer processingImage;
     @GuardedBy("this")
     private FrameMetadata processingMetaData;
-    private ArrayList<Pair<Bitmap, Long>> bufferImageList = new ArrayList<>();
-    private Integer depth = 30;
+
+    class bufferImage
+    {
+        Bitmap image;
+        ImageInfo info;
```
mwarton, line 96: Any reason you are saving the info as well as the bitmap?  You don't seem to be using it in the code.

aquatiko: Yes, it will be used to get rotation degree of image viz. used in text recoginition.
Working locally on it, that's why it's not used here.

vision-quickstart/app/src/main/java/com/google/mlkit/vision/demo/VisionProcessorBase.java
aquatiko: `BitmapUtils` was already present in the boilerplate. I just didn't notice the existence of such function. Found about it after looking into it's code.

PR #4, opened 2020-06-25
SLATE main app

aquatiko:
The app which I'm writing from scratch.


aquatiko:
Till this stage, I've got the cameraX preview working and I can get/access the camera frames in the backend(as required in our application).
Further work: set up text-recogonizer and get it processing on these frames.
Ref:
Screenshot_2020-06-26-13-42-46-95

PR #1, opened 2020-06-15
[EXPERIMENTAL] OCRexperiment: process all frames in a queue

aquatiko:
Main changes(lines) from the boilerplate is added with a comment(per line) : added/changed by rokus
**Main area of interest(files) are:** [CameraSource.java ](https://github.com/aquatiko/screen-latency-testing/blob/97b07cc4b19149a27b8e85b6db2c90019f5e2c21/ocr-reader-complete/app/src/main/java/com/google/android/gms/samples/vision/ocrreader/ui/camera/CameraSource.java#L1), [OCRCaptureActivity.java](https://github.com/aquatiko/screen-latency-testing/blob/97b07cc4b19149a27b8e85b6db2c90019f5e2c21/ocr-reader-complete/app/src/main/java/com/google/android/gms/samples/vision/ocrreader/OcrCaptureActivity.java#L1), [OCRDetectorProcessor.java](https://github.com/aquatiko/screen-latency-testing/blob/97b07cc4b19149a27b8e85b6db2c90019f5e2c21/ocr-reader-complete/app/src/main/java/com/google/android/gms/samples/vision/ocrreader/OcrDetectorProcessor.java#L1)

## Details about experiment
**Currently the app shows:** FrameID, Frame Capture Time (from app start in ms), Detected Text in frame, Time when result was show on screen(after processing everything) 

**Key points noted**: The average delay between two consecutive frame capture is around ~ 40-50 ms on 60fps. It's due to processing load as each frame is processed in this approach as and when it gets captured.

PR #5, opened 2020-07-06
Video analyser

aquatiko:
The main aim of this app is to analyze(i.e do OCR on) videos taken at high fps.

## Current features
- Allows users to select any video from thier storage.
- Load the video in backend and provide an alalysis window giving access to each video frame.
- Do OCR on frames in that window.

## ToDo
- [x] Rotation parameter of videos required for OCR is hardcoded for now because avilable rotation metadata is giving weird/incorrect OCR.
- [x] Total number of frames read at a time is hardcoded for now as accessing a frame id viz. not present gives meomory read errors.
- [x] Able to do OCR on 100 frames only in a session. Find the reason and fix this constraint.

## Main files of interest
From videoAnalyser
[MainActivity.java](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-ce9ab72c6602e37de118a7629012e9f1)
[activity_main.xml](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-690c15ceb6336040cf26ec505cbcc93b)

From Camera2SlowMotion
[CameraFragment.kt](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-688168251bd22bedafbb200c80e769da)
[PermissionsFragment.kt](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-1b4fd01602ea4ed7af651019f699706e)
[SelectorFragment.kt](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-b41b6c60c516a1b90227e696269bd990)
[activity_camera.xml](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-72c9bd32246124fa50c8c26254f48549)
[fragment_camera.xml](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-af2f711f3d15eacfd7d69001bf93651b)

## Future steps
Integrate this with [Camera2SlowMotion](https://github.com/android/camera-samples/tree/master/Camera2SlowMotion) and replace video selection from storage with on demand video capture and processing.


aquatiko:
> Rotation parameter of videos required for OCR is hardcoded for now because avilable rotation metadata is giving weird/incorrect OCR.

I took 4 videos at 4 possible rotation of 0, 90, 180, 270 degree and when they are stored in SD card, they are automatically reoriented at 0 degree and there is no need of handling rotation later in the analyser part.
The OCR can be given a hardcoded rotation of 0 degree for the OCR to work.

aquatiko:
> Able to do OCR on 100 frames only in a session. Find the reason and fix this constraint.

This was a flake error. Working fine now.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -81,14 +82,12 @@ public void onClick(View view) {
                     e.getCause();
                 }
 
-                rotationOfVideo = mediaMetadataRetriever.extractMetadata(MediaMetadataRetriever.METADATA_KEY_VIDEO_ROTATION);
-                List<Bitmap> frameList = mediaMetadataRetriever.getFramesAtIndex(100, 100);
+                totalFrames = Integer.valueOf(mediaMetadataRetriever.extractMetadata(METADATA_KEY_VIDEO_FRAME_COUNT));
+                List<Bitmap> frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames-safeFrames));
```
aquatiko: `safeFrames` are required here because while recording a video at high fps, there are cases when the last frames are not taken correctly.
Eg: Consider a video of time 4.606 seconds taken at 120fps, there will be 553 frames as suggested in metadata.
But when we do the math, we find that there should be 552.72 frames and the last frames is not correctly sotred.

To account this, `safeFrames` value trims some frames from the end.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -0,0 +1,132 @@
...
+            public void onClick(View view) {
+                fileIntent = new Intent(Intent.ACTION_GET_CONTENT);
+                fileIntent.setType("*/*");
+                startActivityForResult(fileIntent, 10);
+            }
+
+        });
+
+        analyzeBtn.setOnClickListener(new View.OnClickListener() {
+            @RequiresApi(api = Build.VERSION_CODES.P)
```
patrickmeiring: Nit: Just curious, where does this come from/why do we need this annotation?

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -0,0 +1,132 @@
...
+import static android.media.MediaMetadataRetriever.METADATA_KEY_VIDEO_FRAME_COUNT;
+
+public class MainActivity extends AppCompatActivity {
+    private TextView filePath;
+    private Button filePickerBtn;
+    private Button analyzeBtn;
+    private Intent fileIntent;
+    private MediaMetadataRetriever mediaMetadataRetriever;
+    private Uri fileUri;
+    private  Integer totalFrames=0;
```
patrickmeiring: Nit: Please run the formatter over this file. I think this should be "private Integer totalFrames = 0".
Can see some strange whitespace issues further down the file (extra lines before end of functions/anonymous class declarations, odd indenting, etc.).

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -0,0 +1,132 @@
...
+
+                totalFrames = Integer.valueOf(mediaMetadataRetriever.extractMetadata(METADATA_KEY_VIDEO_FRAME_COUNT));
+                List<Bitmap> frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames-safeFrames));
+
+                for(int i=0; i<frameList.size();i++)
+                {
+                    // analysis will be done here (forking in a separate thread)
+                    imageHolder = InputImage.fromBitmap(frameList.get(i), 0);
+                    final int finalI = i;
+                    Task<Text> result =
```
patrickmeiring: Observation: Right now this is an experiment, but when we get to implementing this for real, I think it makes sense to track/manage the number of asynchronous tasks you have going on at once to avoid exhausting system resources. This may occur naturally in a binary search implementation anyway, as the next frame to analyse depends on the result of the earlier frame.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -0,0 +1,132 @@
...
+        filePickerBtn = findViewById(R.id.fileSelectButton);
+        analyzeBtn = findViewById(R.id.analyze);
+        recognizer = TextRecognition.getClient();
+
+        filePickerBtn.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                fileIntent = new Intent(Intent.ACTION_GET_CONTENT);
+                fileIntent.setType("*/*");
+                startActivityForResult(fileIntent, 10);
```
patrickmeiring: Please define a constant for '10', so we know it is the same value we are switching on down in 'onActivityResult'.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -0,0 +1,132 @@
...
+                                                    // Task failed with an exception
+                                                    // ...
+                                                }
+                                            });
+                }
+            }
+        });
+    }
+
+    @SuppressLint("MissingSuperCall")
```
patrickmeiring: nit: Any reason we don't also call into the superclass onActivityResult method here?
super.onActivityResult(requestCode, resultCode, data)

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
patrickmeiring: Thanks for the explanation. From your explanation, it sounds like safeFrames = 1 should work (unless a bunch of frames are stored together due to compression?) but I'm not too worried about this since I understand we're trying to do the analysis without storing the video at all.

patrickmeiring (commented):
Thanks for committing, Rohit!

Just a few nits, but otherwise LGTM. Look forward to seeing the analysis integrated with camera capture!

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: Yeah, I also thought `safeFrames = 1` should be enough but for some `safeFrames = 2` was required else I was getting memory leaks. So just to be on safe side I added it as `10` for now.
Thinking now, I realise that this could be the reason.
>(unless a bunch of frames are stored together due to compression?)

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: Sure, thanks for pointing!

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: Interesting idea! I will patch that

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: It was compiler suggested. Tells programmer that an android Pie(9) API is needed.
I think that would limit the number of devices, I will look into this more.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -0,0 +1,134 @@
...
+                                            results.add(visionText.getText());
+                                            Log.d("Rokus logs", "Text detected at index:" + finalI + " " + visionText.getText());
+                                        }
+                                    })
+                                    .addOnFailureListener(
+                                            new OnFailureListener() {
+                                                @Override
+                                                public void onFailure(@NonNull Exception e) {
+                                                    // Task failed with an exception
+                                                    // ...
```
mwarton: Please at least log the exception, nothing worse than code failing silently and driving youself mad trying to find what the problem is.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -0,0 +1,134 @@
...
+                    final int finalI = i;
+                    Task<Text> result =
+                            recognizer.process(imageHolder)
+                                    .addOnSuccessListener(new OnSuccessListener<Text>() {
+                                        @Override
+                                        public void onSuccess(Text visionText) {
+                                            // Task completed successfully
+                                            // ...
+                                            results.add(visionText.getText());
+                                            Log.d("Rokus logs", "Text detected at index:" + finalI + " " + visionText.getText());
```
mwarton: The level of indentation here is getting a bit extreme, can you maybe move the bulk of this code to an analyze method on the main class and just have the onClick handler call that method?

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
mwarton: We will also need a way to get the results back into the main thread.

Most likely have worker threads with a queue of tasks in one direction and results in the other direction.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: The function `getFramesAtIndex` at Line 86 reuires this API level.

`List<Bitmap> frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames-safeFrames))`

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: It seems like I accidently deleted it and added a supress warning on compiler suggestion. Added in refactoring commit.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: Done in refactoring commit.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: Done in refactoring commit.

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
aquatiko: Done in refactoring commit.

aquatiko:
I integrated everything till `refactoring` commit into [Camera2SlowMotion](https://github.com/android/camera-samples/tree/master/Camera2SlowMotion).

The files of interest are:

From videoAnalyser
[MainActivity.java](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-ce9ab72c6602e37de118a7629012e9f1)
[activity_main.xml](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-690c15ceb6336040cf26ec505cbcc93b)

From Camera2SlowMotion
[CameraFragment.kt](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-688168251bd22bedafbb200c80e769da)
[PermissionsFragment.kt](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-1b4fd01602ea4ed7af651019f699706e)
[SelectorFragment.kt](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-b41b6c60c516a1b90227e696269bd990)
[activity_camera.xml](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-72c9bd32246124fa50c8c26254f48549)
[fragment_camera.xml](https://github.com/googleinterns/screen-latency-testing/pull/5/files#diff-af2f711f3d15eacfd7d69001bf93651b)

patrickmeiring:
@aquatiko It looks like you have committed a bunch of IntelliJ caches and generated files. Note that we do not want to track these in source control as they clutter the repository (please only include source code, build scripts and basic project config). Can you amend your last commit to remove the files we don't want and add a .gitignore file that ignores these files for future (like we did in #1)? :)

I will take a look at your second commit when you have done this (the review pane is currently stalling for me with 1700 files).

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -57,14 +58,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
         recognizer = TextRecognition.getClient();
+        mediaMetadataRetriever = new MediaMetadataRetriever();
 
         filePickerBtn.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
                 fileIntent = new Intent(Intent.ACTION_GET_CONTENT);
                 fileIntent.setType("*/*");
-                startActivityForResult(fileIntent, 10);
+                startActivityForResult(fileIntent, filePickerRequestCode);
```
mwarton: Can we name this FILE_PICKER_REQUEST_CODE to make it clear that it is a constant?

patrickmeiring: (Styleguide reference: https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names )

patrickmeiring: ^ Assume that is the styleguide we want you to follow. And if we make comments inconsistent with that guide, feel free to push back :)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -0,0 +1,488 @@
...
+                    recorder.stop()
+
+                    // Removes recording animation
+                    overlay.removeCallbacks(animationTask)
+
+                    // Broadcasts the media file to the rest of the system
+                    MediaScannerConnection.scanFile(
+                            view.context, arrayOf(outputFile.absolutePath), null, null)
+                    Log.d("Rokus Logs:", "reached intent call in kotlin")
+                    // Launch external activity via intent to play video recorded using our provider
```
aquatiko: This functionality from the boilerplate was suppressed and I added the following code to connect both the parts. i.e by using an `intent` to another activity.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -107,24 +107,24 @@ private void analyze(){
...
+                                    Log.d("Rokus logs", "Text detected at index:" + finalI + " " + visionText.getText());
+                                }
+                            })
+                            .addOnFailureListener(
+                                    new OnFailureListener() {
+                                        @Override
+                                        public void onFailure(@NonNull Exception e) {
+                                            // Task failed with an exception
+                                            e.printStackTrace();
+                                            e.getCause();
```
patrickmeiring: Left over from debugging? Throwable.getCause() has no side-effects and the result is unused, so it makes no sense to call it here.

Camera2SlowMotion/utils/src/main/java/com/example/android/camera/utils/AutoFitSurfaceView.kt
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
+        super.onMeasure(widthMeasureSpec, heightMeasureSpec)
+        val width = MeasureSpec.getSize(widthMeasureSpec)
+        val height = MeasureSpec.getSize(heightMeasureSpec)
+        if (aspectRatio == 0f) {
+            setMeasuredDimension(width, height)
+        } else {
+
```
patrickmeiring: nit: Surplus whitespace. (May have been like this in the sample code already.)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
patrickmeiring: Noted, thanks. For review and code history purposes, it would be nice if you could separate your changes from the base sample app.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -0,0 +1,488 @@
...
+//                    startActivity(Intent().apply {
+//                        action = Intent.ACTION_VIEW
+//                        type = MimeTypeMap.getSingleton()
+//                                .getMimeTypeFromExtension(outputFile.extension)
+//                        val authority = "${BuildConfig.APPLICATION_ID}.provider"
+//                        data = FileProvider.getUriForFile(view.context, authority, outputFile)
+//                        flags = Intent.FLAG_GRANT_READ_URI_PERMISSION or
+//                                Intent.FLAG_ACTIVITY_CLEAR_TOP
+//                    })
```
patrickmeiring: Prefer to delete rather than comment out -- if you separate out the base app from your changes, source control will track the deletion and we will always be able to recover the code if we need it by going through the history.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -0,0 +1,488 @@
...
+                    }
+                    catch (exc: Throwable) {
+                        // handler
+                        Log.d("Rokus Logs:", exc.message)
+                        Log.d("Rokus Logs:", exc.cause.toString())
+                        Log.d("Rokus Logs:", exc.stackTrace.toString())
+                    }
+
+
+
```
patrickmeiring, line 364: nit: Extra whitespace here, can you clean this up?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -0,0 +1,488 @@
...
+                    // Finishes our current camera screen
+                    delay(CameraActivity.ANIMATION_SLOW_MILLIS)
+                    navController.popBackStack()
+                }
+            }
+
+            true
```
patrickmeiring, line 371: Don't think this should be here. As above -- might be easier for you and me to pick little things like this up if you can separate your deltas from the base sample app.

patrickmeiring (commented):
Thanks Rohit :) I've made a few comments. Is it possible for you to separate your changes from the base/sample app?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
patrickmeiring: Still seeing this in revision 6da0c5. Looks like this was actually in the sample code: https://github.com/android/camera-samples/blob/master/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt#L379
Can you clean it up in the commit with your changes that follows? :)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
patrickmeiring: nit: I don't think this file should be changed in 6da0c5 (since it is the import of sample code), if you must change the indentation maybe do it in the later commit (or fix the earlier commit). Also, I think the comment about e.getCause() is still outstanding.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -0,0 +1,137 @@
...
+                                    results.add(visionText.getText());
+                                    Log.d("Rokus logs", "Text detected at index:" + finalI + " " + visionText.getText());
+                                }
+                            })
+                            .addOnFailureListener(
+                                    new OnFailureListener() {
+                                        @Override
+                                        public void onFailure(@NonNull Exception e) {
+                                            // Task failed with an exception
+
```
patrickmeiring: nit: Random blank line here

videoanalyser/app/src/main/java/com/example/videoanalyser/MainActivity.java
```diff
@@ -1,14 +1,132 @@
...
+
+        analyzeBtn.setOnClickListener(new View.OnClickListener() {
+            @RequiresApi(api = Build.VERSION_CODES.P)
+            @Override
+            public void onClick(View view) {
+                try {
+                    analyze();
+                }catch (Exception e){
+                    e.printStackTrace();
+                    e.getCause();
```
patrickmeiring: nit: Missed deleting an e.getCause() here. Like the other one you removed, I don't think this there is any point to calling this method as it has no side effects and the result is not used.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.P)
+    private void analyze(){
+        fileUri = Uri.parse(filePath.getText().toString());
+        mediaMetadataRetriever.setDataSource(getApplicationContext(), fileUri);
+        totalFrames = Integer.valueOf(mediaMetadataRetriever.extractMetadata(METADATA_KEY_VIDEO_FRAME_COUNT));
+        List<Bitmap> frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames-safeFrames));
+
+        for(int i=0; i<frameList.size();i++) {
```
patrickmeiring: nit: Still seeing some strange formatting here and elsewhere. Can you run the formatter over this and the other files you changed?

patrickmeiring:
Looking a lot better! Great work :)

LGTM % nits. Just a few issues that were pointed out earlier (formatting, e.getCause()) that still need to be addressed everywhere.

patrickmeiring (commented):
Looking a lot better! Great work :)

LGTM % nits. Just a few issues that were pointed out earlier (formatting, e.getCause()) that still need to be addressed everywhere.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
mwarton: can you set up either format on save or format on commit?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -38,7 +38,7 @@
 public class MainActivity extends AppCompatActivity {
     private Integer totalFrames=0;
```
mwarton: I think these two should be final and CONSTANT_CASE two right?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -359,16 +357,21 @@ class CameraFragment : Fragment() {
...
-                        flags = Intent.FLAG_GRANT_READ_URI_PERMISSION or
-                                Intent.FLAG_ACTIVITY_CLEAR_TOP
-                    })
+                    // Launch analyser activity via intent
+                    val intent = Intent(view.context, MainActivity::class.java)
+                    val authority = "${BuildConfig.APPLICATION_ID}.provider"
+                    intent.putExtra("file URI", FileProvider.getUriForFile(view.context, authority, outputFile).toString())
+                    Log.d("Rokus Logs:", "starting intent call in kotlin")
+                    try {
+                        // some code
```
mwarton: delete useless comments?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
aquatiko: `safeFrames` can be constant but I'm setting `totalFrames` from metadata later on. Initialized here as I didn't wanted a garbage value to indicate that `totalFrames` is set.

mwarton (approved):
LGTM

PR #6, opened 2020-07-14
App support for server automation

aquatiko:
Adding support in the app to handle server(laptop) automating the workflow.

The end goal should be an app which can communicate with server and start/stop it's activities in it's lifecycle and finally send the results i.e lag tests.

Tempory server script: [Server scripts](https://gist.github.com/aquatiko/ca7754942fe0b59b0a622b263313dfa4)
## Current features
- Server can find a random avilable port for communication.
- Server can open app automatically on android while sending it's binded port information to app.
- Server can set-up a tunnel for android and laptop to communicate viz. reverse port forwarding using adb and the port info, which it sent
- App can send a signal to laptop when capture button is pressed to start simulating keystrokes on laptop end
- Server can receive that signal and start to simulate keystrokes

## ToDo
- [x] Convert server scripts to `golang` and use `tast`.
- [x] Automate app activity till capture button press. (Currently user is required to select a fps and then press capture on the app)
- [x] Add singnal to (in app and server) stop recording when all inputs are simulated.
- [x] Send simulation info. to app.
- [x] Calculate results (lag tests).


Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -39,6 +62,11 @@ class CameraActivity : AppCompatActivity() {
             container.systemUiVisibility = FLAGS_FULLSCREEN
         }, IMMERSIVE_FLAG_TIMEOUT)
     }
+ // TODO
+//    override fun onDestroy() {
```
mwarton: Please don't leave commented code in commits.

A Todo comment is fine.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -17,18 +17,41 @@
...
+    })
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_camera)
         container = findViewById(R.id.fragment_container)
+        val intent = intent
+        val bundle = intent!!.extras
+        laptopPort = bundle?.getString("port")?.toInt() ?: 0
+        if (laptopPort != 0) {
```
mwarton, line 50: what if it fails?  Should we log something or have some other communication so the server can retry?

(note, the answer can be no :) )

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -17,18 +17,41 @@
...
+        } catch (e: IOException) {
+            e.printStackTrace()
+        }
+    })
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_camera)
         container = findViewById(R.id.fragment_container)
+        val intent = intent
```
mwarton: why do we need a local variable to shadow the member?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -54,6 +54,7 @@ protected void onCreate(Bundle savedInstanceState) {
     analyseResultField = findViewById(R.id.analyseResultView);
 
     analyseResultField.setMovementMethod(new ScrollingMovementMethod());
+    analyseResultField.append("Port binded:" + CameraActivity.Companion.getLaptopPort());
```
mwarton: This should probably be "Port bound"

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -17,18 +17,41 @@
...
+import java.io.BufferedReader
+import java.io.IOException
+import java.io.InputStreamReader
+import java.io.PrintWriter
+import java.net.Socket
 
 class CameraActivity : AppCompatActivity() {
 
     private lateinit var container: FrameLayout
+    private var SERVER_IP = "127.0.0.1"
```
mwarton: This looks a little strange as it is actually the IP of the phone.  I assume this is because you are port forwarding, but maybe the SERVER_IP name is not the best.  Its ok to leaveas I don't really have a better suggestion.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -308,7 +308,8 @@ class CameraFragment : Fragment() {
                     // Prevents screen rotation during the video recording
                     requireActivity().requestedOrientation =
                             ActivityInfo.SCREEN_ORIENTATION_LOCKED
-
+                    CameraActivity.output.write("started capture")
```
mwarton: We probably want something a little more structured than raw text.  Proto is probably best, but JSON would be much easier to set up and almost as good.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
aquatiko: This was done on purpose as the `SERVER_IP` can be changed to a network IP too, and the same implementation will work. Here currenlty I'm using a localhost becasue of port-forwarding.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
aquatiko: Will fix in the following commits

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -308,7 +308,7 @@ class CameraFragment : Fragment() {
                     // Prevents screen rotation during the video recording
                     requireActivity().requestedOrientation =
                             ActivityInfo.SCREEN_ORIENTATION_LOCKED
-                    CameraActivity.output.write("started capture")
+                    CameraActivity.output.write("started capture*")
```
aquatiko: `*` acts as a delimiter here. (Based on current golang server script, will be improved in JSON formatting of messages)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
mwarton: I think you will need to change a lot more than just a hard-coded IP address if you change to connecting to the server by IP.

Also, this should be "const val" rather than var if you are treating it as a constant (CAPITAL_CASE)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -104,35 +111,100 @@ private void analyze() {
...
+    frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames - SAFE_FRAMES));
+
+    AsyncTaskServer asyncTaskServer = new AsyncTaskServer();
+    asyncTaskServer.execute();
+
+    AsyncTaskOCR asyncTaskOCR = new AsyncTaskOCR();
+    asyncTaskOCR.execute();
+  }
+
+  private class AsyncTaskServer extends AsyncTask<String, String, String> {
```
mwarton: I don't understand what this code is doing, can you please add a comment or choose more descriptive names?

It seems to be receiving a series of timestamps from the server, but I'm not clear on what these are timestamps of or what you intend to do with them.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -230,13 +234,31 @@ private void parseServer() {
...
   }
+
+  private void showResults() {
+    if (deltaServer.size() < 2 || deltaOCR.size() < 2) {
+      Log.d(tag, "No results to show");
+      return;
+    }
+    Log.d("Show Results", "Video start timestamp:" + CameraFragment.Companion.getRecordingStartTime());
+    String serverChar = "";
+    Integer sync_offset = deltaOCR.get(0) - convertStringTimeToInt(CameraFragment.Companion.getRecordingStartTime());
```
aquatiko, line 254: `deltaOCR[0]` have the starting timestamp when tests are begun on laptop. After 2 seconds of that, the keys are smulated.

Server/go.sum
```diff
@@ -0,0 +1,45 @@
+github.com/BurntSushi/freetype-go v0.0.0-20160129220410-b763ddbfe298/go.mod h1:D+QujdIlUNfa0igpNMk6UIvlb6C252URs4yupRUV4lQ=
```
aquatiko, line 1: [ Need reviews] Not sure if this file needs to be commited as well.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -38,6 +37,7 @@
   private final Integer FILE_PICKER_REQUEST_CODE = 10;
   private String tag = "Rokus Logs:";
   private Integer fps;
+  private Long sync_offset;
```
mwarton: This should be camelCase

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -66,7 +65,7 @@ protected void onCreate(Bundle savedInstanceState) {
     analyseResultField = findViewById(R.id.analyseResultView);
 
     analyseResultField.setMovementMethod(new ScrollingMovementMethod());
-    analyseResultField.append("\nPort bound:" + CameraActivity.Companion.getLaptopPort());
+    analyseResultField.append("\nPort bound:" + CameraActivity.Companion.getLaptopPort() +"\n");
```
mwarton: can you just have a displayLine() method or something to make it clear that this is being displayed on the screen vs sent over the network

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -144,7 +143,10 @@ protected String doInBackground(String... strings) {
 
         while (message != null) {
           message = CameraActivity.input.readLine();
-          if (message != null) serverCache.add(message);
+          if (message == null) {
```
mwarton: Can we please move all of the network communication to its own dedicated class.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -144,7 +143,10 @@ protected String doInBackground(String... strings) {
...
+            break;
+          }
+          serverCache.add(message);
```
mwarton: can this please be named something more descriptive?  It is not a cache.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -162,7 +164,9 @@ protected void onPostExecute(String s) {
         parseServer();
       } catch (Exception e) {
         Log.d(tag, "Parsing server cache failed");
-        Log.d(tag, String.valueOf(e.getCause()));
+        Log.d("Rokus Logs:", String.valueOf(e.getCause()));
```
mwarton: Any reason for the "Rokus logs"?  The tag should be something related to the app,not the author.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -205,60 +211,45 @@ public void onFailure(@NonNull Exception e) {
...
-        if (!last.equals(resultsOCR.get(i))) {
-          last = resultsOCR.get(i);
-          tmpOCR.add(i);
-          deltaOCR.add(recordStartTime + (i * offset));
-          Log.d(tag, "In parseOCR:" + i + "  " + (recordStartTime + (i * offset)));
-        }
+      Long recordStartTime = CameraFragment.Companion.getRecordingStartMillis();
+      Long offset = 1000L / fps;
+      for (int i = 0; i < resultsOCR.size(); i++) {
+        deltaOCR.add(recordStartTime + (i * offset));
```
mwarton: if deltaOCR is just this simple calculation, why do we need to store the values?  just have a function.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -205,60 +211,45 @@ public void onFailure(@NonNull Exception e) {
...
   }
 
-  // Parse timestamps received from server. Last 12 characters of time is important here.
+  // Parse timestamps received from server.
   private void parseServer() {
-    Integer length = serverCache.get(0).length() - 12;
     for (int i = 0; i < serverCache.size(); i++) {
-      String tmp = serverCache.get(i).substring(length);
-      deltaServer.add(convertStringTimeToInt(tmp));
+      deltaServer.add(Long.valueOf(serverCache.get(i)));
```
mwarton: You should do the Long conversion when you receive them over the network so that any failures are closer to the source of the error.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -205,60 +211,45 @@ public void onFailure(@NonNull Exception e) {
...
     } catch (Exception e) {
       Log.d("Rokus Logs:", "Parsing OCR failed with: " + e.getMessage());
     }
   }
 
-  // Parse timestamps received from server. Last 12 characters of time is important here.
+  // Parse timestamps received from server.
   private void parseServer() {
-    Integer length = serverCache.get(0).length() - 12;
     for (int i = 0; i < serverCache.size(); i++) {
```
mwarton: optional: you could use a map here instead.  Both ways are fine, but if you don't know about map in Java you should learn.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -205,60 +211,45 @@ public void onFailure(@NonNull Exception e) {
...
+    for (int j = 1; j < deltaServer.size(); j++) {
+      for (int i = 0; i < resultsOCR.size(); i++) {
+        if (serverSequence.equals(resultsOCR.get(i))) {
+          analyseResultField.append(
+              "Lag:" + (deltaOCR.get(i) - deltaServer.get(j) + sync_offset)
+                  + " Server sequence:" + serverSequence + "\n");
+          break;
+        }
+      }
+      serverSequence += "m";
```
mwarton, line 219: the server text to look for should be sent from the server along with the timestamps, not recalculated in the app.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -162,7 +164,9 @@ protected void onPostExecute(String s) {
         parseServer();
       } catch (Exception e) {
         Log.d(tag, "Parsing server cache failed");
-        Log.d(tag, String.valueOf(e.getCause()));
+        Log.d("Rokus Logs:", String.valueOf(e.getCause()));
+        Log.d("Rokus Logs:", e.getMessage());
```
mwarton: Can't you just Log.d("SLATE", "Parsing server message failed", e); ?

Why do you print the exception fields explicitly?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -188,63 +165,53 @@ public void onSuccess(Text visionText) {
...
-  private void parseOCR() {
+  private void setFrameTimeStamp() {
     try {
-      Long recordStartTime = CameraFragment.Companion.getRecordingStartMillis();
-      Long offset = 1000L / fps;
-      for (int i = 0; i < resultsOCR.size(); i++) {
-        deltaOCR.add(recordStartTime + (i * offset));
-      }
+      videoFrameTimestamp.add(recordStartTime + (framesProcessed * frameDuration));
     } catch (Exception e) {
```
mwarton, line 191: ArrayList.add doesn't throw any exceptions...

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -188,63 +165,53 @@ public void onSuccess(Text visionText) {
...
-        "Video start timestamp:" + CameraFragment.Companion.getRecordingStartTime());
+        "Video start timestamp:" + recordStartTime);
+    //TODO: Server sequence is hardcoded. Add functionality to receive serverSequence.
     String serverSequence = "m";
-    sync_offset = deltaServer.get(0) - CameraFragment.Companion.getRecordingStartMillis();
-    Log.d("Show Results", "Sync offset:" + sync_offset);
-    for (int j = 1; j < deltaServer.size(); j++) {
+    syncOffset = serverTimestamp.get(0) - CameraFragment.Companion.getRecordingStartMillis();
+    Log.d("Show Results", "Sync offset:" + syncOffset);
+    for (int j = 1; j < serverTimestamp.size(); j++) {
```
mwarton, line 210: I feel like this timestamp data and related calculations should be moved to a separate class so that it can be grouped together more obviously and not distributed through all of the activity and task scheduling code in this class.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -188,63 +165,53 @@ public void onSuccess(Text visionText) {
...
     Log.d(
         "Show Results",
-        "Video start timestamp:" + CameraFragment.Companion.getRecordingStartTime());
+        "Video start timestamp:" + recordStartTime);
+    //TODO: Server sequence is hardcoded. Add functionality to receive serverSequence.
     String serverSequence = "m";
-    sync_offset = deltaServer.get(0) - CameraFragment.Companion.getRecordingStartMillis();
-    Log.d("Show Results", "Sync offset:" + sync_offset);
-    for (int j = 1; j < deltaServer.size(); j++) {
+    syncOffset = serverTimestamp.get(0) - CameraFragment.Companion.getRecordingStartMillis();
```
mwarton, line 208: Is it possible to send this in the result of the activity rather than using an (effectively global) variable to communicate it?

Server/go.sum
mwarton: https://github.com/golang/go/wiki/Modules#releasing-modules-all-versions

Ensure your go.sum file is committed along with your go.mod file.

Server/serverSlate.go
```diff
@@ -0,0 +1,84 @@
...
+	timeStamps := make([]string, KEY_PRESS_COUNT+1)
+
+	message, _ := hostReader.ReadString('*') // using * as delim and hardcoded in app to append * at end of message
+	fmt.Println(message)
+	if message == "started capture"+"*" {
+		simulateKeyPress(KEY, KEY_PRESS_COUNT, timeStamps)
+	}
+	fmt.Println("Key simulation ended")
+
+	message, _ = hostReader.ReadString('*') // using * as delim and hardcoded in app to append * at end of message
```
mwarton, line 49: don't ignore errors.

Server/serverSlate.go
```diff
@@ -0,0 +1,84 @@
...
+
+func main() {
+	fmt.Println("Start server...")
+
+	ln, _ := net.Listen("tcp", "127.0.0.1:")
+	_, serverPort, _ := net.SplitHostPort(ln.Addr().String())
+	fmt.Println("Listening on address:")
+	fmt.Println(ln.Addr())
+
+	openAppCommand := exec.Command("adb", "shell", "am", "start", "-n",
```
mwarton, line 26: put this in an appropriately named function

Server/serverSlate.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/go-vgo/robotgo"
+)
+
+const KEY = "m"
+const KEY_PRESS_COUNT = 10
+
+func main() {
+	fmt.Println("Start server...")
+
+	ln, _ := net.Listen("tcp", "127.0.0.1:")
```
mwarton, line 21: move all the network communication and message parsing into a separate class (possibly in another file)

Server/serverSlate.go
```diff
@@ -0,0 +1,84 @@
...
+	err := cmd.Run()
+	if err != nil {
+		fmt.Println(fmt.Sprint(err) + ": " + stderr.String())
+		return 0
+	}
+	fmt.Println("Result: " + out.String())
+	return 1
+}
+
+func simulateKeyPress(key string, keyPressCount int, timeStamps []string) {
```
mwarton, line 70: this should be simulateKerPresses as it is doingmultiple.

mwarton, line 70: this should return timestamps, not take it as an input parameter that is modified.

You should also return errors that occur during execution.

Server/serverSlate.go
```diff
@@ -0,0 +1,84 @@
...
+	timeStamps[0] = strconv.Itoa(int(time.Now().UnixNano() / 1000000))
+	time.Sleep(1 * time.Second)
+	for i := 1; i <= keyPressCount; i++ {
+		robotgo.TypeStr(key)
```
mwarton, line 74: I assume this returns an error, don't ignore it.

mwarton:
As discussed we will merge this pull request as the experimental code. REmaining comments will be addressed in subsequent pull requests.

PR #7, opened 2020-07-28
Add missing utils files

aquatiko:
These were part of the Camera2SlowMotion sample app which is required to run the project.
They exisited on my local project but was shadowed by a .gitignore while pushing to master.

This fix will enable people to run the app from the master branch without any warnings/errors.


Camera2SlowMotion/.gitignore
```diff
@@ -1 +0,0 @@
-utils/
```
aquatiko, line 1: This was shadowing the missing `utils` folder.

mwarton (approved):

PR #8, opened 2020-08-01
Remove user interaction in activity

aquatiko:
Previously user was shown all the available high speed camera setting to
chose from. 
This is bypassed with automatic selection of lowest workable setting.

**Note:** The lowest workable setting has camera fps greater or equal to 120. 
The HighSpeedCameraConstraint API make sure of this as any fps lower than 120 fps is not treated as high speed fps.


aquatiko:
The first activity in which user can select camera setting is automated and the camera capture activity will be shown directly.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/SelectorFragment.kt
```diff
@@ -57,15 +58,11 @@ class SelectorFragment : Fragment() {
...
-            adapter = GenericListAdapter(cameraList, itemLayoutId = layoutId) { view, item, _ ->
-                view.findViewById<TextView>(android.R.id.text1).text = item.title
-                view.setOnClickListener {
-                    Navigation.findNavController(requireActivity(), R.id.fragment_container)
-                            .navigate(SelectorFragmentDirections.actionSelectorToCamera(
-                                    item.cameraId, item.size.width, item.size.height, item.fps))
-                }
-            }
+            // select lowest resolution and lowest high speed fps setting available
+            val item = cameraList.get(cameraList.size-1);
```
mwarton: Is the ordering gauranteed here?  Can you leave a comment pointing to some documentation?

mwarton: item doesn't make sense as a name anymore, it is no longer a list item.  Maybe make this setting or resolution?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/SelectorFragment.kt
```diff
@@ -36,6 +36,7 @@ import com.example.android.camera2.slowmo.R
 /**
  * In this [Fragment] we let users pick a camera, size and FPS to use for high
  * speed video recording
+ * Update: This part has been automated to select a lowest working setting for SLATE.
  */
 class SelectorFragment : Fragment() {
```
mwarton: You shouldn't need the fragment at all any more right?  Can you simplify this down and remove the fragment?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/SelectorFragment.kt
```diff
@@ -36,6 +36,7 @@ import com.example.android.camera2.slowmo.R
 /**
  * In this [Fragment] we let users pick a camera, size and FPS to use for high
  * speed video recording
+ * Update: This part has been automated to select a lowest working setting for SLATE.
```
mwarton: Please just change the whole comment, don't make readers read a change log.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/SelectorFragment.kt
aquatiko: I realised that there's a bug which is making the app crash on some chosen settings. This means that automating the selection of the last setting from the list will not work. I'm chasing that bug. This PR will be **blocked** till then.
Upon that I will simplify down this fragment.

**Ref:** [TODO](https://github.com/googleinterns/screen-latency-testing/blob/d934ba1b444d3076ff2c255146d4774fb8149d4f/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java#L67), #10

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/SelectorFragment.kt
aquatiko: Sure

patrickmeiring:
I've reviewed, but have nothing to add beyond what Matthew has said. My main question is now that the selection is automated, why do we retain the UI control?

aquatiko:
> Is the ordering gauranteed here? Can you leave a comment pointing to some documentation?

Added a function `findLowestCameraSetting` to check and select the lowest camera settings from all available settings.

> You shouldn't need the fragment at all any more right? Can you simplify this down and remove the fragment?

Merged `Permissions` and `Selector` fragments into one thereby removing the `selector` UI.

Ready to be merged after #11

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
```diff
@@ -16,70 +16,96 @@
...
+            // Request camera-related permissions
+            requestPermissions(PERMISSIONS_REQUIRED, PERMISSIONS_REQUEST_CODE)
+        }
+    }
+
+    /** Selects the lowest camera fps and the lowest image capture resolution available among them.*/
+    private fun findLowestCameraSetting(cameraList: List<CameraInfo>): CameraInfo {
+        var bestLowestSettingSeen = cameraList.get(0)
+        for (setting in cameraList) {
+            if (setting.fps <= bestLowestSettingSeen.fps &&
```
mwarton: probably want to check fps is at least 60.

mwarton: Do we need to prioritize between these?

eg if we have:
1. 240 fps * 1080
2. 129 fps * 1920

This algorithm will only choose whichever is first in the list.  Maybe we should make a rough heuristic as to how long OCR will take?  I imagine this is FPS * (some constant * width * height)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
```diff
@@ -16,70 +16,96 @@
...
+    override fun onRequestPermissionsResult(
+            requestCode: Int, permissions: Array<String>, grantResults: IntArray) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults)
+        if (requestCode == PERMISSIONS_REQUEST_CODE) {
+            if (grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+                // Takes the user to the success fragment when permission is granted
+                Navigation.findNavController(requireActivity(), R.id.fragment_container).navigate(
+                        PermissionsFragmentDirections.actionPermissionsFragmentToCameraFragment2(lowestSetting.cameraId, lowestSetting.size.width, lowestSetting.size.height, lowestSetting.fps))
+            } else {
+                Toast.makeText(context, "Permission request denied", Toast.LENGTH_LONG).show()
```
mwarton: Please expand the error message.  

"Permission request denied.  You must grant permissions for this app to function. Please restart the app and grant permissions."

mwarton: Leave a TODO to report this error back to the server.  If it is running unattended in the lab, no-one will be looking at the screen to see the error, we need the error to show up on the server so we can get someone to go and grant the permissions.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
```diff
@@ -16,70 +16,96 @@
...
     }
 
     companion object {
 
+        /** Convenience method used to check if all permissions required by this app are granted */
+        fun hasPermissions(context: Context) = PERMISSIONS_REQUIRED.all {
+            ContextCompat.checkSelfPermission(context, it) == PackageManager.PERMISSION_GRANTED
+        }
+
+        private lateinit var lowestSetting: CameraInfo
```
mwarton: Rather than storing this as a member, please abstract a new method for detecting the lowest setting and navigating to the other activity immediately.  

Also, I think this should be stored on the object itself, not the companion object.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
```diff
@@ -16,70 +16,96 @@
...
+
         private data class CameraInfo(
                 val title: String,
                 val cameraId: String,
                 val size: Size,
                 val fps: Int)
 
         /** Converts a lens orientation enum into a human-readable string */
-        private fun lensOrientationString(value: Int) = when(value) {
+        private fun lensOrientationString(value: Int) = when (value) {
```
mwarton: formatting error?

Camera2SlowMotion/app/src/main/res/navigation/nav_graph.xml
```diff
@@ -22,27 +22,11 @@
...
-        android:id="@+id/selector_fragment"
-        android:name="com.example.android.camera2.slowmo.fragments.SelectorFragment"
-        android:label="Selector" >
-
         <action
-            android:id="@+id/action_selector_to_camera"
-            app:launchSingleTop="true"
-            app:destination="@id/camera_fragment" />
-
+          android:id="@+id/action_permissions_fragment_to_camera_fragment"
```
mwarton, line 28: probably don't need the "fragment" here, the others didn't use it before.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
aquatiko: @mwarton The high speed camera API don't consider 60 as a high enough fps. Anything up and above 120 fps will be shown in the list.
Thus, 120fps or greater is guranteed if the phone supports the setting.

Camera2SlowMotion/app/src/main/res/navigation/nav_graph.xml
aquatiko: The name suggest the link between `permission_fragment` and `camera_fragment`. There was another UI panel in this diagram before which I removed viz. `selector_fragment`.
Thus, this new action replaces previous links from `permission_fragment` -> `selector_fragment`  -> `camera_fragment` with `permission_fragment` -> `camera_fragment`
Screenshot 2020-08-05 at 9 00 04 AM

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
aquatiko: Done.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
aquatiko: Yes, running the Google java linter on the whole file highlighted this  for the boilerplate app code.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
aquatiko: Done.
> Please abstract a new method for detecting the lowest setting and navigating to the other activity immediately.

This can't be done at the moment because in this fragment, the UI will be waiting for the permission. If it already has that, it moves with the paramaters of the `lowestSetting`

Currenlty this fragment is intented for asking permissions only and I migrated code from selector fragment. I'm planning to migrate this to AnalyserFragment(new) when merging of CameraActivity and AnalyserActivity happens. 

Until then I'm logically moving code around while combining selector and permissions fragment.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
aquatiko: Fixed.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
aquatiko: Done.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsAndCameraSettingFragment.kt
mwarton: I think maybe you misunderstood...

Make a new function:

```
private fun startCamera() {
       val cameraManager =
                requireContext().getSystemService(Context.CAMERA_SERVICE) as CameraManager
        val cameraList = enumerateHighSpeedCameras(cameraManager)
        val lowestSetting = findLowestCameraSetting(cameraList)

        Navigation.findNavController(requireActivity(), R.id.fragment_container).navigate(
                    PermissionsFragmentDirections.actionPermissionsToSelector())
                    PermissionsFragmentDirections.actionPermissionsFragmentToCameraFragment(
                            lowestSetting.cameraId, lowestSetting.size.width, lowestSetting.size.height, lowestSetting.fps))
}
```

And call it in the two places you currently have similar code.

That way there is no need to store lowestSetting.

aquatiko: Ah, I see.
Fixed in the commit. Thanks for pointing!

mwarton (approved):

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsFragment.kt
```diff
@@ -31,34 +35,65 @@ private val PERMISSIONS_REQUIRED = arrayOf(
...
+
+        val lowestSetting = findLowestCameraSetting(cameraList)
+
+        Navigation.findNavController(requireActivity(), R.id.fragment_container).navigate(
+                PermissionsFragmentDirections.actionPermissionsFragmentToCameraFragment(
+                        lowestSetting.cameraId, lowestSetting.size.width, lowestSetting.size.height, lowestSetting.fps))
+    }
+    /** Selects the lowest camera fps and the lowest image capture resolution available among them.*/
+    private fun findLowestCameraSetting(cameraList: List<CameraInfo>): CameraInfo {
+        var bestLowestSettingSeen = cameraList.get(0)
```
patrickmeiring: optional: Wonder if you can do this more simply with a Stream:
https://developer.android.com/reference/java/util/stream/Stream
and/or a comparator. Something like:
Comparator<CameraInfo> comparator = Comparator.comparing(s -> s.fps);
comparator = comparator.thenComparing(Comparator.comparing(s -> s.size.width));
comparator = comparator.thenComparing(Comparator.comparing(s -> s.size.height));
return cameraList.stream().sorted(comparator).findFirst().get(); // Alternative is to copy the cameraList (to avoid modifying the original collection, which the caller might not expect) and use .sort() directly on the List.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsFragment.kt
```diff
@@ -31,34 +35,65 @@ private val PERMISSIONS_REQUIRED = arrayOf(
...
+        val cameraList = enumerateHighSpeedCameras(cameraManager)
+
+        val lowestSetting = findLowestCameraSetting(cameraList)
+
+        Navigation.findNavController(requireActivity(), R.id.fragment_container).navigate(
+                PermissionsFragmentDirections.actionPermissionsFragmentToCameraFragment(
+                        lowestSetting.cameraId, lowestSetting.size.width, lowestSetting.size.height, lowestSetting.fps))
+    }
+    /** Selects the lowest camera fps and the lowest image capture resolution available among them.*/
+    private fun findLowestCameraSetting(cameraList: List<CameraInfo>): CameraInfo {
```
patrickmeiring: Do we also need to filter by orientation (i.e. back facing?)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsFragment.kt
```diff
@@ -31,34 +35,65 @@ private val PERMISSIONS_REQUIRED = arrayOf(
...
 
+    private fun startCamera(){
+        val cameraManager =
+                requireContext().getSystemService(Context.CAMERA_SERVICE) as CameraManager
+
+        val cameraList = enumerateHighSpeedCameras(cameraManager)
+
+        val lowestSetting = findLowestCameraSetting(cameraList)
+
+        Navigation.findNavController(requireActivity(), R.id.fragment_container).navigate(
```
patrickmeiring: Any reason why we find the camera options here in the permissions fragment, rather than in the Camera fragment?

patrickmeiring (commented):
Thanks, this looks like a useful change. Comments below.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsFragment.kt
aquatiko: Ah, I missed that. Thanks for pointing.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/PermissionsFragment.kt
aquatiko: That makes more sense. As discussed I will move this logic to CameraFragment.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -481,5 +499,61 @@ class CameraFragment : Fragment() {
...
+                                .REQUEST_AVAILABLE_CAPABILITIES_CONSTRAINED_HIGH_SPEED_VIDEO)) {
+                    // For each camera, list its compatible sizes and FPS ranges
+                    cameraConfig.highSpeedVideoSizes.forEach { size ->
+                        cameraConfig.getHighSpeedVideoFpsRangesFor(size).forEach { fpsRange ->
+                            val fps = fpsRange.upper
+                            val info = CameraInfo(
+                                    "$orientation ($id) $size $fps FPS", id, size, fps)
+
+                            // Only report the highest FPS in the range, avoid duplicates. Select only Back cameras for SLATE.
+                            if (!availableCameras.contains(info) && orientation.equals("Back")) availableCameras.add(info)
```
aquatiko, line 549: added check to fliter non-back camera orientations.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -62,8 +60,8 @@ import kotlin.coroutines.suspendCoroutine
 
 class CameraFragment : Fragment() {
 
-    /** AndroidX navigation arguments */
-    private val args: CameraFragmentArgs by navArgs()
+    /** Lowest Camera setting for recording video */
+    private lateinit var lowestSetting: CameraInfo
```
patrickmeiring: Maybe rename to "cameraSetting" and have the comment say "Camera setting to use when recording video".
While it is also the lowest available setting, for the purposes of most of the other code, the important thing is that this stores setting we are using to record video, not that it is the lowest available setting.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -460,6 +460,24 @@ class CameraFragment : Fragment() {
         recorderSurface.release()
     }
 
+    /** Selects the lowest camera fps and the lowest image capture resolution available among them.*/
+    private fun findLowestCameraSetting(cameraList: List<CameraInfo>): CameraInfo {
+        var bestLowestSettingSeen = cameraList.get(0)
```
patrickmeiring, line 465: optional: See earlier comment regarding using streams/Comparators here.

patrickmeiring (commented):
LGTM % nits

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
aquatiko: okay, makes sense.

patrickmeiring (approved):

PR #9, opened 2020-08-04
Code Refactor

aquatiko:
Upon completion of the prototyping phase, this PR aims to refactor the experimental code-base.


Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+    videoProcessor = new VideoProcessor();
+
+    lagCalculator = new LagCalculator(videoProcessor, CameraActivity.serverHandler);
+
+    resultPublisher = new ResultPublisher();
+
+    analyseResultField.append(
+        "\nPort bound:" + CameraActivity.serverHandler.getServerSocketPort() + "\n");
+
+    CameraFragment.Companion.getFilePath();
```
mwarton: Why do you call this twice?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+    filePath.setText(CameraFragment.Companion.getFilePath());
+
+    filePickerBtn.setOnClickListener(
+        new View.OnClickListener() {
```
mwarton: Are you able to change to using lambdas for these?  There is a lot of low information content boilerplate code here.

I haven't personally done it before, but for example: https://mayojava.github.io/android/java/using-java8-lambda-expressions-in-android/

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+        });
+
+    analyzeBtn.setOnClickListener(
+        new View.OnClickListener() {
+          @RequiresApi(api = Build.VERSION_CODES.P)
+          @Override
+          public void onClick(View view) {
+            try {
+              analyze();
+            } catch (Exception e) {
```
mwarton: Are you expecting any particular exceptions here?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+        });
+  }
+
+  @Override
+  protected void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
+    super.onActivityResult(requestCode, resultCode, data);
+    if (requestCode == FILE_PICKER_REQUEST_CODE && resultCode == RESULT_OK) {
+      String path = data.getData().toString();
+      fileUri = data.getData();
+      filePath.setText(path);
```
mwarton: filePath.setText(fileUri.toString()); ?  Then you can remove the first line and only call getData once.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  /** Triggers subsequent lifecycle events of lag calculations. */
+  @RequiresApi(api = Build.VERSION_CODES.P)
+  private void analyze() {
+    fileUri = Uri.parse(filePath.getText().toString());
```
mwarton: Why is fileUri a member variable if we overwrite it every time we call analyze()?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+    videoProcessor.setVideoReader(getApplicationContext(), fileUri);
```
mwarton: A setVideoReader function should take a VideoReader argument, perhaps createVideoReader is a beter name?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+
+    CameraActivity.serverHandler.downloadServerTimeStamps();
```
mwarton: Can we make ServerHandler a singleton rather than piggybacking off of the CameraActivity singleton?

It doesn't really make sense when reading the code why this needs to reference CameraActivity.  It should at least be commented, but ideally this file shouldn't need to reference CameraActivity at all.  I believe Activities in Android should be independent and only communicate by Intents.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+
+    videoProcessor.doOcrOnVideo(lagCalculator);
```
mwarton: Just doOcr ?  Also should it be doOCR? not sure what the Java style here is.

mwarton: lagCalculator already knows about the videoProcessor (passed in the constructor), why do you need to make them mutually dependant?  try to make your dependencies go only one way (dependency graph should be acyclic)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
mwarton: Would probably be better if this method returned the timestamps and you passed them as a parameter to where they are needed.  When you do it this way you are storing state in the object and it may not be obvious when something doesn't work that it is because this function wasn't called.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /** A background threads which listens for the results of LagCalculator and shows the results
+   * when available to TextView.*/
+  private class ResultPublisher extends AsyncTask<LagCalculator, Void, Void> {
+    private Integer resultPublishedCount = 0;
+
+    @Override
+    protected Void doInBackground(LagCalculator... lagCalculators) {
+      while (true) {
+        if (resultPublishedCount < lagCalculators[0].lagResult.size()) {
```
mwarton: You are only using the 0th element:
1. How do you know the list isn't empty?
2. Why aren't you using the rest?

If it is only a single object, please try to pass a single object rather than an array of size 1, and if not possible, please document why it is not possible, and the answers to the above two points.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+          analyseResultField.append(
+              "Char = "
```
mwarton: use String.format instead.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
mwarton: Having a loop counter as a class member variable is a pretty far from normal coding construct.  Please make it local or document clearly why it needs to be this way.

mwarton (commented):
I've only looked at the first file, probably more to come tomorrow

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.activity_analyser);
+
+    filePath = findViewById(R.id.selectedFilePath);
+    filePickerBtn = findViewById(R.id.fileSelectButton);
+    analyzeBtn = findViewById(R.id.analyze);
+    analyseResultField = findViewById(R.id.analyseResultView);
+
+    analyseResultField.setMovementMethod(new ScrollingMovementMethod());
```
patrickmeiring: I'm somewhat partial to doing this all in XML if possible, and leaving the code behind to business logic as much as possible. How about wrapping your TextView in a ScrollView and removing this?
Various options are discussed here: https://stackoverflow.com/questions/1748977/making-textview-scrollable-on-android

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
patrickmeiring: Agree. In this version of the code, I'd probably vote to remove fileUri as a class-level variable and keep only one copy of state (namely in the textbox). Down the track, when we integrate capture and analysis, and delete the file picker, we can switch back to using a fileUri if need be.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
patrickmeiring: I personally wouldn't bother with the Singleton if the intent is to merge CameraActivity & AnalyserActivity.

While I don't like the way the code is now (activities should only communicate with intents), the issue existed in the old code (MainActivity was using the sockets on CameraActivity directly) and this would all become local to the activity when CameraActivity & AnalyserActivity are merged. Maybe put in a TODO though.

patrickmeiring: Agree. Prefer to make the data flow obvious through method results/parameters, not flow in a hidden way.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
patrickmeiring: Similar comments on the class structure, but I think we should go one step further. 
Either:
- Encapsulate videoProcessor inside lagCalculator (or vica-versa), or 
- Make videoProcessor and lagCalculator independent objects (that don't reference eachother) and let the AnalyserActivity pass results from one to the other.

At the moment, it looks like videoProcessor is used from both inside lagCalculator and in AnalyserActivity. This feels like AnalyserActivity is forming part of the implementation of lagCalculator (i.e. lagCalculator fails to abstract AnalyserActivity from its implementation details; we must read both classes together to understand how either work).

If AnalyserActivity must delegate videoProcessor and lagCalculator to communicate, prefer to use an interface that makes clear what communication that is occurring and decouples LagCalculator from VideoProcessor.

Overall, the flow that seems most logical to me right now is:

Timestamps inputTimestamps = serverHandler.getInputTimestamps();
OcrResults videoTimestamps = VideoProcessor.getVideoTimestamps(video);
LagResults inputLags = LagCalculator.computeLags(inputTimestamps, videoTimestamps);

Where Timestamps/OcrResults/Lags are only container classes that hold data, and VideoProcessor and LagCalculator are static classes (have no state).

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /** A background threads which listens for the results of LagCalculator and shows the results
+   * when available to TextView.*/
+  private class ResultPublisher extends AsyncTask<LagCalculator, Void, Void> {
+    private Integer resultPublishedCount = 0;
+
+    @Override
+    protected Void doInBackground(LagCalculator... lagCalculators) {
+      while (true) {
```
patrickmeiring: Why not make this a for loop? That seems simpler and would make loop termination clearer.
for (int resultPublishedCount = 0; resultPublishedCount < ... .size(); resultPublishedCount++) {
  ...
}
You could probably then also rename resultPublishedCount to 'i'.

patrickmeiring (commented):
Good work, just taking a break from reviewing as well as it is getting late. 

To help us, could you also provide a statement of what this refactoring is (and is not?) intending to do?

aquatiko:
Working on merging `AnalyserActivity` and `CameraActivity`. 
That will fix some of the code-reviews by simplifying the logic and rest will be fixing along with it too.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,50 @@
...
+    Log.d(ContentValues.TAG, "Server-Host Sync offset:" + serverHostSyncOffset);
+
+    ArrayList<Long> serverTimestamp = serverData.getServerTimeStamps();
+    ArrayList<Long> videoFrameTimestamp = videoData.getVideoFrameTimestamp();
+    ArrayList<String> resultsOCR = videoData.getResultsOCR();
+
+    // Finds the video frame with the running serverSequence text.
+    for (int j = 1; j < serverTimestamp.size(); j++) {
+      for (int i = 0; i < resultsOCR.size(); i++) {
+        if (serverSequence.equals(resultsOCR.get(i))) {
```
patrickmeiring: It seems like this may not behave as expected if two characters were to appear on the screen at the same time. Thoughts about perhaps doing prefix matching?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,111 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.PrintWriter;
+import java.net.Socket;
+import java.util.ArrayList;
+
+/** Handles every interaction of app with the server. */
+public class ServerHandler {
+
+  private Integer serverSocketPort;
```
patrickmeiring, line 18: Prefer 'int' over Integer here, as there is no need for this to be nullable. (Also update corresponding getters and setters.)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,111 @@
...
+  private final String SERVER_IP = "127.0.0.1";
+  private ArrayList<Long> serverTimeStamps = new ArrayList<>();
+  private Long serverStartTimeStamp;
```
patrickmeiring: Ditto, prefer the unboxed type here.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,111 @@
...
+  private ConnectionThread connectionThread;
+  private Socket socket;
+  public BufferedReader inputReader;
+  public PrintWriter outputWriter;
+  private Long hostSyncTimeStamp;
```
patrickmeiring: Ditto (long instead of Long)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,111 @@
...
+/** Handles every interaction of app with the server. */
+public class ServerHandler {
+
+  private Integer serverSocketPort;
+  private final String SERVER_IP = "127.0.0.1";
+  private ArrayList<Long> serverTimeStamps = new ArrayList<>();
+  private Long serverStartTimeStamp;
+  private ConnectionThread connectionThread;
+  private Socket socket;
+  public BufferedReader inputReader;
```
patrickmeiring: Why is this field public? If you need to expose the socket reader/writer, prefer to write a getter that returns them, and let external users do: serverHandler.getInputReader().read(...). 
I don't think external users of this class should be able to set this field.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    return serverTimeStamps;
+  }
+
+  public Long getServerStartTimeStamp() {
+    return serverStartTimeStamp;
+  }
+
+  public void startConnection() {
+    connectionThread = new ConnectionThread();
+    connectionThread.execute();
```
patrickmeiring, line 47: It seems like we never wait for this task to complete. How does the consumer of this class know when it is safe to call downloadServerTimeStamps()? Do we need a join or await here?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,111 @@
...
+  private final String SERVER_IP = "127.0.0.1";
+  private ArrayList<Long> serverTimeStamps = new ArrayList<>();
+  private Long serverStartTimeStamp;
+  private ConnectionThread connectionThread;
+  private Socket socket;
+  public BufferedReader inputReader;
+  public PrintWriter outputWriter;
+  private Long hostSyncTimeStamp;
+
+  public Long getHostSyncTimeStamp() {
```
patrickmeiring: Timestamp is one word, so this should be "Timestamp" not "TimeStamp"
https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

(please do find and replace through this class, I see many occurances.)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.List;
+
+/**
+ * Loads recorded video file into media-reader, attach timestamps to individual video frames and
+ * requests Ocr on each frame. Upon Ocr processing of all frames it requests LagCalculator class to
+ * further compute the lag results.
+ */
+public class VideoProcessor {
+
+  private static final Integer SAFE_FRAMES = 10;
```
patrickmeiring: Please to use 'int' over "Integer'; there is no need for this to be nullable (ditto for occurances below)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /** Convenience method used to assign frame timestamps to individual frames based on video
+   * recording start time and the fps of video. */
+  private void setNextFrameTimeStamp() {
+    videoFrameTimestamp.add(recordStartTime + (framesProcessed * frameDuration));
+  }
+
+  public void doOcrOnVideo(LagCalculator lagCalculator) {
+    AnalyseVideo ocrOnVideoTask = new AnalyseVideo();
+    ocrOnVideoTask.execute(lagCalculator);
```
patrickmeiring: How does the client know this task has completed and the results are available? You're not awaiting or returning a handle to the AsyncTask that is returned.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+    return recordStartTime;
+  }
+
+  public ArrayList<Long> getVideoFrameTimestamp() {
+    return videoFrameTimestamp;
+  }
+
+  private Long recordStartTime = CameraFragment.Companion.getRecordingStartMillis();
+  private Long frameDuration = 1000L / CameraFragment.Companion.getFpsRecording();
+  private Integer framesProcessed = 0;
```
patrickmeiring: Ditto x3

patrickmeiring (commented):
Added some more comments. Some of the comments may be a little bit superfluous as they relate to preexisting issues (so we can defer addressing them until later).

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/AnalyserActivity.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -16,41 +16,32 @@
...
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_camera)
         container = findViewById(R.id.fragment_container)
+
         val bundle = intent!!.extras
-        laptopPort = bundle?.getString("port")?.toInt() ?: 0
+        val laptopPort = bundle?.getString("port")?.toInt() ?: 0
         if (laptopPort != 0) {
-            connectionThread.start()
+            serverHandler.serverSocketPort = laptopPort
```
mwarton: It would be better to pass this in the constructor for serverHandler.  Is there any reason you are constructing it before onCreate?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -63,6 +54,19 @@ class CameraActivity : AppCompatActivity() {
         }, IMMERSIVE_FLAG_TIMEOUT)
     }
 
+    internal fun analyse(fileUri: Uri){
```
mwarton: probably better to use the american spelling: analyze

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -63,6 +54,19 @@ class CameraActivity : AppCompatActivity() {
...
+        val videoProcessor = VideoProcessor()
+        val lagCalculator = LagCalculator(videoProcessor, serverHandler)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            videoProcessor.createVideoReader(applicationContext, fileUri)
+        };
+
+        serverHandler.downloadServerTimeStamps()
```
mwarton: can you return and pass the timestamps as an argument?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -63,6 +54,19 @@ class CameraActivity : AppCompatActivity() {
         }, IMMERSIVE_FLAG_TIMEOUT)
     }
 
+    internal fun analyse(fileUri: Uri){
+        val videoProcessor = VideoProcessor()
+        val lagCalculator = LagCalculator(videoProcessor, serverHandler)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
mwarton: What if this is not true?  Shouldn't an error be raised or something?

Does continuing on in this scenario make sense?

Can we just mark the App as requiring this build version and remove this (and any other similar) check?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,51 @@
...
+import android.util.Log;
+import java.util.ArrayList;
+
+/**
+ * Calculates lag of each key-press on the server using the key-press timestamps from server and
+ * character seen timestamp from recorded video.
+ */
+public class LagCalculator {
+  private VideoProcessor videoData;
+  private ServerHandler serverData;
```
mwarton: maybe call it serverHandler rather than serverData, it is more than just some data.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,51 @@
...
+import android.content.ContentValues;
+import android.util.Log;
+import java.util.ArrayList;
+
+/**
+ * Calculates lag of each key-press on the server using the key-press timestamps from server and
+ * character seen timestamp from recorded video.
+ */
+public class LagCalculator {
+  private VideoProcessor videoData;
```
mwarton: maybe call it videoProcessor

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,51 @@
...
+  private ServerHandler serverData;
+  public ArrayList<Long> lagResult = new ArrayList<Long>();
```
mwarton: array names should be plural: lagResults

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public LagCalculator(VideoProcessor videoData, ServerHandler serverData) {
+    this.videoData = videoData;
+    this.serverData = serverData;
+  }
+
+  public ArrayList<Long> calculateLag() {
```
mwarton: This should be a static method that takes all the data it needs as parameters.  There is no need for any of the class variables here, they can only introduce subtle bugs later.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,51 @@
...
+    ArrayList<Long> serverTimestamp = serverData.getServerTimestamps();
+    ArrayList<Long> videoFrameTimestamp = videoData.getVideoFrameTimestamp();
+    ArrayList<String> resultsOCR = videoData.getResultsOCR();
+
+    // Finds the video frame with the running serverSequence text.
+    for (int j = 1; j < serverTimestamp.size(); j++) {
+      for (int i = 0; i < resultsOCR.size(); i++) {
+        if (resultsOCR.get(i).startsWith(serverSequence)) {
+          long lag = videoFrameTimestamp.get(i) - serverTimestamp.get(j) + serverHostSyncOffset;
+          lagResult.add(lag);
```
mwarton: Since this is a member variable, if you call this function twice you will get the results from both calls returned in the second call.  Please make lagResult a local variable only.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,51 @@
...
+          Log.d("Lag:", String.valueOf(lag));
+          break;
+        }
+      }
+      serverSequence += "m";
```
mwarton, line 39: Please leave a todo to let the server inform the app what text to be expecting each time rather than hardcoding it.  (doesn't need to fix in this PR)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,51 @@
...
+    Log.d(ContentValues.TAG, "Video start timestamp:" + videoData.getRecordStartTime());
+    // TODO: Server sequence is hardcoded. Add functionality to receive serverSequence.
+    String serverSequence = "m";
+    long serverHostSyncOffset =
+        serverData.getServerStartTimestamp() - serverData.getHostSyncTimestamp();
+    Log.d(ContentValues.TAG, "Server-Host Sync offset:" + serverHostSyncOffset);
+
+    ArrayList<Long> serverTimestamp = serverData.getServerTimestamps();
+    ArrayList<Long> videoFrameTimestamp = videoData.getVideoFrameTimestamp();
+    ArrayList<String> resultsOCR = videoData.getResultsOCR();
```
mwarton: How is multiple text detected by OCR in a single frame handled?  We need to be resilient to this in the final app.  Doesn't need to be in this PR, but please start thinking about it.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
aquatiko: Already added!

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
aquatiko: Using `@RequiresApi(Build.VERSION_CODES.P)` will make sure to add the same to app API level requirment.
Thanks for pointing. Will add this in subsequent commit.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
aquatiko: No specific reason. Will fix this.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -16,41 +16,34 @@
...
-        try {
-            socket = Socket(SERVER_IP, laptopPort)
-            output = PrintWriter(socket.getOutputStream())
-            input = BufferedReader(InputStreamReader(socket.getInputStream()))
-            Log.d(TAG_AUTHOR, "Connection established")
-        } catch (e: IOException) {
-            e.printStackTrace()
-        }
-    })
+    internal lateinit var serverHandler: ServerHandler
```
mwarton: can this be private?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,81 @@
...
+          .get();
+    } catch (ExecutionException | InterruptedException e) {
+      e.printStackTrace();
+      return null;
+    }
+  }
+
+  private class LagCalculatorThread extends AsyncTask<LagCalculatorPrams, Void, ArrayList<Long>> {
+    @Override
+    protected ArrayList<Long> doInBackground(LagCalculatorPrams... params) {
```
mwarton: Why are you doing this in the background?  This step is just a few hundred string comparisons right?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,81 @@
...
+      ArrayList<Long> serverTimestamps,
+      ArrayList<Long> videoFrameTimestamp,
+      ArrayList<String> resultsOcr,
+      long syncOffset) {
+    LagCalculatorThread lagCalculatorThread = new LagCalculatorThread();
+    try {
+      return lagCalculatorThread
+          .execute(
+              new LagCalculatorPrams(serverTimestamps, videoFrameTimestamp, resultsOcr, syncOffset))
+          .get();
```
mwarton: calling get blocks the main thread waiting for the result, ie it is exactly the same as calling this directly without putting it in a separate task, as mentioned in the other comment, I don't think anything in this class needs to be async.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -0,0 +1,81 @@
...
+public class LagCalculator {
+
+  private static class LagCalculatorPrams {
+    ArrayList<Long> serverTimestamps, videoFrameTimestamps;
+    ArrayList<String> resultsOcr;
+    long syncOffset;
+
+    public LagCalculatorPrams(
+        ArrayList<Long> serverTimestamps,
+        ArrayList<Long> videoFrameTimestamps,
```
mwarton: I think you should create a data structure that holds both the timestamp and the text, and have a single array.  This will reduce the number of params and should also make the code simpler. (no need for indexes in your calculation loop)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
aquatiko: fixed

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  public ArrayList<Long> downloadServerTimeStamps() {
+    if (serverSocketPort == null || !socket.isConnected()) {
+      Log.d(ContentValues.TAG, "Can't find server to download timestamps. No port bound or connection is dead.");
+      return null;
+    }
+    DownloadServerLogs downloadServerLogsTask = new DownloadServerLogs();
+    try {
+      return downloadServerLogsTask.execute().get();
```
mwarton, line 87: again, this is blocking, so there is no reason to put the code in another thread.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,124 @@
...
+    return serverStartTimestamp;
+  }
+
+  public void startConnection() {
+    ConnectionThread connectionThread = new ConnectionThread();
+    connectionThread.execute();
+  }
+
+  public void sendKeySimulationSignal() {
+    hostSyncTimestamp = System.currentTimeMillis();
```
mwarton, line 52: leave a todo for a more robust time sync measurement (we need at least a few round trips averaged)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -0,0 +1,124 @@
...
+      ArrayList<Long> serverTimestamps = new ArrayList<>();
+      try {
+        Log.d(ContentValues.TAG, "Sending request to download timestamps");
+        outputWriter.write("send timestamps" + "*");
+        outputWriter.flush();
+        Log.d(ContentValues.TAG, "Request sent to server. Waiting to receive timestamps..");
+        String message = inputReader.readLine();
+        serverStartTimestamp = Long.parseLong(message);
+
+        while (message != null) {
```
mwarton, line 113: does this mean the connection is closed after the last timestamp?  How do you send the results back to the server?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        ocrTasks.add(
+            recognizer
+                .process(imageHolder)
+                .addOnSuccessListener(
+                    visionText -> {
+                      resultsOCR.add(visionText.getText());
+                      Log.d(
+                          ContentValues.TAG,
+                          "Text detected at index:" + finalI + " " + visionText.getText());
+                      setNextFrameTimeStamp();
```
mwarton: This is a major concurrency problem.  You are calling this function which assumes it is called on frames in sequential order, however you are calling it from an async completion function which can be run for the frames out of order. In addition the accesses to member variables would need to be locked by mutex to prevent corruption.  You need to preserve the knowledge of which frame is doing this.

Also, is there any reason to do this here?  It could easily be done in the non-async part of the code right?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /** Convenience method used to assign frame timestamps to individual frames based on video
+   * recording start time and the fps of video. */
+  private void setNextFrameTimeStamp() {
+    videoFrameTimestamp.add(recordStartTime + (framesProcessed * frameDuration));
```
mwarton: This is accessing 4 member variables, many of which seem to only be member variables to enable this function.  Please either pass the variables in/out, or get rid of this helper.

mwarton (commented):
Maybe leave the threading stuff as is and we will do another pull request.  I think we probably need to move to something more robust than AsyncTask, like CompleteableFuture.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
aquatiko: I'm keeping this non-private becasue `camera_fragment`(viz. an associated fragment to this CameraActivity) needs to use `serverHandler` and making it private will not allow the fragment to access it's own activity object.

aquatiko: Ref [L#290](https://github.com/googleinterns/screen-latency-testing/blob/e4acf3f854f780729be639a644f29c2d592f1aa2/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt#L290)  [L#332](https://github.com/googleinterns/screen-latency-testing/blob/e4acf3f854f780729be639a644f29c2d592f1aa2/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt#L332)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: @mwarton No, the connection is not closed until explicitly done so, or until the `serverHandler` object goes out of scope.
Here it only means that until the inputReader is able to get messages, store that message and read one line at a time from the socket.

mwarton: I guess I'm not clear on why it would return null instead of blocking for more data if the connection is still open.  I assumed it was blocking until the connection was closed by the other side.

Seems odd, but ok for now.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
aquatiko: Moved the member variables to function scope.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
aquatiko: Shifted to non-async part

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: done

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: will address in a seperate PR

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
aquatiko: This is simplified without an async-task in the following commit.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
aquatiko: Simplified in a non-async part.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
aquatiko: Simplified in a non-async part.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
aquatiko: Note that the timestamps data is sent from server in a continous stream and this inputReader tries to read it line by line i.e delimiting on `\n` and at the end of that continous stream if would read a null giving us cue that the timestamps are received completely.
Also, I tested with the setup, worked as expected.

mwarton (approved):

PR #11, opened 2020-08-04
Fix app crash on video frame read

### LINKED ISSUE #10: App crashing on some video recording settings
## Expected Behavior
The app should work with all video-recording settings in the first selector Activity.

## Actual Behavior
App crashes on some video-recording setting after pressing `START ANALYSIS` button.

## Steps to Reproduce the Problem

1. Install app on specified master version.
1. Try the end-to-end experiement using Server with all settings shown on selector activity.
1. The app will crash on some of the setting.

## Specifications

- Version: App based on commit 1a449d3be6ef1ea5d051b8d93e711cc491d14d25

aquatiko:
Previously app was trying to read all video frames in a single request
to `mediaMetadataRetriever`, now video file is loaded in chunks.

**Fixes:** #10 

Looks like some limitation on `mediaMetadataRetriever` API being unable to read all frames at once.
**Note:** A disadvantage obeserved while reading in chunks was longer load time opposed to a single load requests (which crashes on large requests).

### DIFF
--- a/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
+++ b/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
@@ -120,6 +120,7 @@ public class MainActivity extends AppCompatActivity {
             i, Math.min(totalFrames - i, FRAME_CHUNK_READ_SIZE)));
       }catch (IllegalStateException e){
         Log.d(ContentValues.TAG, "Video chunk read unsuccessful with: "+ e.getMessage());
+        finish();
       }
 
     }
--- a/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
+++ b/Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
@@ -2,6 +2,7 @@ package com.example.android.camera2.slowmo;
 
 import static android.media.MediaMetadataRetriever.METADATA_KEY_VIDEO_FRAME_COUNT;
 
+import android.content.ContentValues;
 import android.content.Intent;
 import android.graphics.Bitmap;
 import android.media.MediaMetadataRetriever;
@@ -33,7 +34,7 @@ import java.util.List;
 public class MainActivity extends AppCompatActivity {
   private Integer totalFrames = 0;
   private Integer framesProcessed = 0;
-  private final Integer SAFE_FRAMES = 10;
+  private final Integer FRAME_CHUNK_READ_SIZE = 100;
   private final Integer FILE_PICKER_REQUEST_CODE = 10;
   private final String TAG_AUTHOR = "Rokus Logs:";
   private Long syncOffset;
@@ -113,8 +114,15 @@ public class MainActivity extends AppCompatActivity {
     totalFrames =
         Integer.valueOf(mediaMetadataRetriever.extractMetadata(METADATA_KEY_VIDEO_FRAME_COUNT));
     // TODO: Add on demand frame read opposed to read all frames.
-    frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames - SAFE_FRAMES));
+    for (int i = 0; i < totalFrames; i += FRAME_CHUNK_READ_SIZE) {
+      try{
+        frameList.addAll(mediaMetadataRetriever.getFramesAtIndex(
+            i, Math.min(totalFrames - i, FRAME_CHUNK_READ_SIZE)));
+      }catch (IllegalStateException e){
+        Log.d(ContentValues.TAG, "Video chunk read unsuccessful with: "+ e.getMessage());
+      }
 
+    }
     DownloadServerLogs downloadServerLogsTask = new DownloadServerLogs();
     downloadServerLogsTask.execute();
 


Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -113,8 +113,12 @@ private void analyze() {
     totalFrames =
         Integer.valueOf(mediaMetadataRetriever.extractMetadata(METADATA_KEY_VIDEO_FRAME_COUNT));
     // TODO: Add on demand frame read opposed to read all frames.
-    frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames - SAFE_FRAMES));
-
+    for (int i = 0; i < totalFrames; i += FRAME_CHUNK_READ_SIZE) {
+      if (i + FRAME_CHUNK_READ_SIZE - 1 < totalFrames)
+        frameList.addAll(mediaMetadataRetriever.getFramesAtIndex(i, FRAME_CHUNK_READ_SIZE));
+      else
+        frameList.addAll(mediaMetadataRetriever.getFramesAtIndex(i, totalFrames-i));
```
mwarton: frameList.addAll(mediaMetadataRetriever.getFramesAtIndex(i, Math.min(totalFrames - i, FRAME_CHUNK_READ_SIZE))

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -113,8 +113,12 @@ private void analyze() {
     totalFrames =
         Integer.valueOf(mediaMetadataRetriever.extractMetadata(METADATA_KEY_VIDEO_FRAME_COUNT));
     // TODO: Add on demand frame read opposed to read all frames.
-    frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames - SAFE_FRAMES));
```
mwarton, line 116: Any details on why this crashed?  did you get an excpetion message?

I would of assumed it was because it was too large for memory, but in that case the new code should crash too, so I'm not sure what the problem is.

mwarton (commented):
Looks ok if it works, but a better design would be to only read chunks of frames as we need them for analysis, and then discard them after the analysis to reduce memory usage.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
aquatiko: That makes sense. Thanks!
> Looks ok if it works, but a better design would be to only read chunks of frames as we need them for analysis, and then discard them after the analysis to reduce memory usage.

I added a TODO to read frames on demand, I planned to do that when optimizing the linear algorithm for finding `serverSequence` among the OCR texts.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
aquatiko: @mwarton Added a log to catch the exception.
Added details to the issue #10 discussion.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -113,8 +114,15 @@ private void analyze() {
...
+    for (int i = 0; i < totalFrames; i += FRAME_CHUNK_READ_SIZE) {
+      try{
+        frameList.addAll(mediaMetadataRetriever.getFramesAtIndex(
+            i, Math.min(totalFrames - i, FRAME_CHUNK_READ_SIZE)));
+      }catch (IllegalStateException e){
+        Log.d(ContentValues.TAG, "Video chunk read unsuccessful with: "+ e.getMessage());
```
mwarton, line 122: Shouldn't we abort the analysis?  Incomplete or completely missing frames will mess up the results won't they?

aquatiko: @mwarton fixed. I added a command to kill the activity and fall back to previous activity to start the capture again.

mwarton (approved):

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
```diff
@@ -113,8 +114,16 @@ private void analyze() {
     totalFrames =
         Integer.valueOf(mediaMetadataRetriever.extractMetadata(METADATA_KEY_VIDEO_FRAME_COUNT));
     // TODO: Add on demand frame read opposed to read all frames.
-    frameList = mediaMetadataRetriever.getFramesAtIndex(0, Math.max(0, totalFrames - SAFE_FRAMES));
+    for (int i = 0; i < totalFrames; i += FRAME_CHUNK_READ_SIZE) {
+      try{
+        frameList.addAll(mediaMetadataRetriever.getFramesAtIndex(
+            i, Math.min(totalFrames - i, FRAME_CHUNK_READ_SIZE)));
+      }catch (IllegalStateException e){
```
patrickmeiring, line 121: Please run the formatter to fix up the whitespace before submitting.
https://google.github.io/styleguide/javaguide.html#s4-formatting
try {
...
} catch (IllegalStateException e) {
 ...
}

patrickmeiring (commented):
LGTM % nits

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/MainActivity.java
patrickmeiring: As discussed, I'll merge now and we'll fix this in a follow up.

PR #13, opened 2020-08-10
Robust threading interface

aquatiko:
This PR aims to replace AsyncTask implementation used to achieve threading with CompletableFuture.


Server/serverSlate.go
```diff
@@ -67,18 +67,22 @@ func runOsCommand(cmd *exec.Cmd) int {
...
 }
 
-func simulateKeyPress(key string, keyPressCount int, timeStamps []string) {
-	timeStamps[0] = strconv.Itoa(int(time.Now().UnixNano() / 1000000))
+func simulateKeyPress(key string, keyPressCount int, timeStamps []string, keyPressSequence []string) {
 	time.Sleep(1 * time.Second)
-	for i := 1; i <= keyPressCount; i++ {
+	runningSeq := ""
+	for i := 0; i < keyPressCount; i++ {
+		androidTime, _ := exec.Command("adb", "shell", "date", "+%s%3N").Output()
```
aquatiko: Previously at this point, I was querrying current server UTC time and storing them. They were later sent to android device and used in lag calculation. As the server clock was not in sync with android, a mechanism to find the drift between two clocks were also implemented.(and used in lag calculation)

In this new implementation, we querry the current UTC time of android device and store them. They are later sent to android device and used in lag calculation. As there is only a single clock now, there is no need to find drift now.

Benchmarking the command, found execution time of command as
```
real    0m0.108s
user    0m0.006s
sys     0m0.014s
```
This can be used as an offset to further sync received androidTime. _(Some discussion is required on this end)_

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -56,22 +57,18 @@ class CameraActivity : AppCompatActivity() {
...
-        }
+        videoProcessor.createVideoReader(applicationContext, fileUri)
 
-        val serverTimestamps = serverHandler.downloadServerTimeStamps()
+        val serverTimestampsAndSequence = serverHandler.downloadServerTimeStampsAndSequence()
 
         val resultsOcr = videoProcessor.doOcr()
 
-        lagCalculator.calculateLag(serverTimestamps, videoProcessor.videoFramesTimestamp, resultsOcr, serverHandler.getSyncOffset())
+        val lagResults = lagCalculator.calculateLag(serverTimestampsAndSequence, videoProcessor.videoFramesTimestamp, resultsOcr)
```
patrickmeiring: It might be just as easy to await the futures by calling .get() out here, and passing LagCalculator the results. That way LagCalculator is purely about calculation and doesn't have to think about threading at all (which may make easier to unit test).

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,110 @@
...
-          message = inputReader.readLine();
-        }
-        Log.d(ContentValues.TAG, "All timestamps from server received.");
-      } catch (IOException e) {
-        Log.d(ContentValues.TAG, "Server TimeStamp read failed");
-      }
-      return serverTimestamps;
-    }
+  @RequiresApi(api = VERSION_CODES.N)
+  public CompletableFuture<ServerData> downloadServerTimeStampsAndSequence() {
```
patrickmeiring: nit: downloadServerTimeStampsAndSequence should be downloadServerTimestampsAndSequence() (TimeStamps should be Timestamps.)

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,110 @@
...
+    return connection.thenApply(
+        conn -> {
+          synchronized (conn) {
+            ServerData serverData = new ServerData();
+            try {
+              Log.d(ContentValues.TAG, "Sending request to download timestamps");
```
patrickmeiring: uber nit: Periods are beng used inconsistently at some of some log messages. Here there is no period, but below 
 you do use one (on "All timestamps from server received.").

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,110 @@
...
+              conn.outputWriter.write("send timestamps" + "*");
+              conn.outputWriter.flush();
+              Log.d(
+                  ContentValues.TAG, "Request sent to server. Waiting to receive timestamps..");
```
patrickmeiring: "Waiting to receive timestamps.." I think this should end with three periods or one.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -32,100 +31,100 @@
 public class VideoProcessor {
```
patrickmeiring, line 33: Some high-level comments. There are appear to be three computations defined in this class:
1. reading video frames (defined in createVideoReader). Inputs: Video file. Output: Frames
2. constructing an array of frame timestamps (setFramesTimestamp). Inputs: Frames (length), Time of first frame, recording FPS. Output: List of timestamps.
3. frame-by-frame OCR (doOcr). Input: frames: Output: text in each frame.

The results of all computations ultimately get consumed by computation 4 (lag analysis). 

You are defining each of 1, 2 and 3 as a seperate future, with 2 and 3 chained on the results of (1). 

Observations:
- There is no real advantage to parallelising (2) and (3) because (2) is incredibly quick. The way you are synchronising (2) and (3) means they don't run in parallel anyway, so I guess this was not your goal to start with.
- The results of (1) only get used in (2) and (3), and (2) and (3) only get used in (4). So why not combine, (1), (2) and (3) together into a single future that can be awaited? 

The inputs of the combined computation would be the video file, time of first frame, recording FPS, and the result would be the OCR results with timestamps. 

Once combined into a single computation, you will be remove all the state (frameList, videoFramesTimestamp) from this class. It would also be pretty simple to extricate the creation of a future to outside this class.

Something like:
```
public ArrayList<FrameAnalysis> process(Uri fileUri, long recordingStartTime, int fps)  {
   frames = readFrames(fileUrl);
   ocrResults = doOcr(frames);
   result = new ArrayList<FrameAnalysis>()
   for (int i = 0; i < ocrResults.size(); i++) {
        result.add(new FrameAnalysis(ocrResults.get(i), timestampForFrame(i, longRecordingStartTime, fps))
   }
   return result
}
```
Note there are currently two ArrayLists, which leaves open the idea they may be of a mismatched length. By combining the results into a single ArrayList (e.g. `ArrayList<FrameAnalysis>`), we make clear from the return type that this can never occur.

- I would also suggest you let the future be created in analyse method of the camera activity (so you can be blind to threading concerns here).

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
```diff
@@ -56,22 +57,18 @@ class CameraActivity : AppCompatActivity() {
...
-    // The results are calculated while the app UI closes. Add a UI until the processing completes.
     @RequiresApi(Build.VERSION_CODES.P)
     internal fun analyze(fileUri: Uri) {
         val videoProcessor = VideoProcessor()
         val lagCalculator = LagCalculator()
 
-        if (!videoProcessor.createVideoReader(applicationContext, fileUri)) {
-            finish()
-        }
+        videoProcessor.createVideoReader(applicationContext, fileUri)
```
patrickmeiring, line 65: Further down, I'm going to suggest you move the creation of all futures to this method, and await them here as well. 

Also, unless it is demonstrated that we really need to parallelise OCR and downloading timestamps from the server, we could consider just running them sequentially as well? calculateLag appears to be blocking until the futures complete, so there seems to be no concern about blocking the UI thread either.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,110 @@
...
-
+  /**
+   * Creates a socket for communication with the server. Sets reader and writer for server
+   * connection.
+   */
+  @RequiresApi(api = VERSION_CODES.N)
   public void startConnection() {
-    ConnectionThread connectionThread = new ConnectionThread();
-    connectionThread.execute();
+    connection =
```
patrickmeiring, line 67: I can't see where startConnection() is used from the changes here, but would it be better to to explicitly wait until the connection was established before we started capture? That way, we can handle any errors and don't start sending keypresses late (if the connection took a while to establish).

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,110 @@
...
+            });
   }
 
-  // TODO: Use round-trip averaging to fix hostSyncTimestamp
+  @RequiresApi(api = VERSION_CODES.N)
   public void sendKeySimulationSignal() {
-    hostSyncTimestamp = System.currentTimeMillis();
-    outputWriter.write("started capture*");
-    outputWriter.flush();
+        connection.thenApply(
```
patrickmeiring, line 88: I think the way you are chaining futures here (and ignoring the result) means that any errors that occurred in the connection stage, or in the this stage, will never be surfaced. I think we should try to find a way for them to bubble up to the top-level control flow.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,110 @@
...
-      } catch (IOException e) {
-        Log.d(ContentValues.TAG, "Server TimeStamp read failed");
-      }
-      return serverTimestamps;
-    }
+  @RequiresApi(api = VERSION_CODES.N)
+  public CompletableFuture<ServerData> downloadServerTimeStampsAndSequence() {
+    return connection.thenApply(
+        conn -> {
+          synchronized (conn) {
```
patrickmeiring, line 106: It seems odd that we're having to synchronise on this conn object here. I guess the concern is that sendKeySimulationSignal() is going to run at the same time as downloadServerTimeStampsAndSequence. 
Maybe we should keep sendKeySimulationSignal() blocking (like the old implementation), so this can never occur?

Server/serverSlate.go
```diff
@@ -38,18 +37,19 @@ func main() {
 
 	hostReader := bufio.NewReader(conn)
 	timeStamps := make([]string, KEY_PRESS_COUNT+1)
+	keyPressSequence := make([]string, KEY_PRESS_COUNT+1)
 
 	message, _ := hostReader.ReadString('*') // using * as delim and hardcoded in app to append * at end of message
 	fmt.Println(message)
 	if message == "started capture"+"*" {
-		simulateKeyPress(KEY, KEY_PRESS_COUNT, timeStamps)
+		simulateKeyPress(KEY, KEY_PRESS_COUNT, timeStamps, keyPressSequence)
```
patrickmeiring: The way this is being passed, it makes it seem like keyPressSequence is an input of this method, when it really is an output (?). Why not make the method create and return the slice instead?

Server/serverSlate.go
patrickmeiring: Thanks Rohit. I think this is an important improvement. 

I have some doubts as to how reliable this synchronisation method will be (longer-term), but let's catch up offline about this -- I know you are already aware of better synchronisation options.

patrickmeiring (changes requested):
Thanks Rohit. I think we are going in the right direction. Comments below.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
mwarton: where does videoProcessor.videoFramesTimestamp get calculated?  It would be better if it was returned from somewhere (either part of resultsOCR or returned from createVideoReader).

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/LagCalculator.java
```diff
@@ -12,31 +17,27 @@
...
-      List<String> resultsOcr,
-      long serverHostSyncOffset) {
+      CompletableFuture<ServerData> serverData,
+      CompletableFuture<ArrayList<Long>> videoFramesTimestamp,
+      CompletableFuture<ArrayList<String>> resultsOcr)
+      throws ExecutionException, InterruptedException {
 
     ArrayList<Long> lagResults = new ArrayList<>();
-    if (serverTimestamps.size() < 2 || videoFrameTimestamp.size() < 2) {
+    if (serverData.get().timestamps.size() < 2 || videoFramesTimestamp.get().size() < 2) {
```
mwarton: won't this block the main thread?
 You should do a allOf until all the needed futures are completed and only use get in the method that runs once they are all complete.  You could also use thenCombine if you only have two futures you need.

Ideally this should be done in the calling function so this class doesn't need to import competablefuture at all.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
mwarton: use exceptionally to catch and log any thrown exceptions at the end of any futures chains that do not return anything.

mwarton, line 88: Use thenAccept if you have nothing to return.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
mwarton: Generally the socket and both streams are not thread safe.  I think it is better to be safe than sorry here, especially given that the implementation may change in the future.

Server/serverSlate.go
mwarton: 100ms execution time means this error from this synchronization is on the same order as what we are trying to measure right?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/CameraActivity.kt
aquatiko: yeah, makes sense! Thanks for pointing.

Server/serverSlate.go
aquatiko: @patrickmeiring @mwarton 
I realised that the command used here is not the most efficient one.
When I enter the android shell using `adb shell` and inside it I can see:
```shell
RMX1971:/ $ time date +%s%3N                                                                  
1597121530314
    0m00.04s real     0m00.02s user     0m00.02s system
```

In previous benchmark the time of opening android shell got added.

aquatiko: As discussed about this approach, due the significant overheads of using adb connection this way, this approach is dropped.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,147 @@
...
+    PrintWriter outputWriter;
   }
 
-  public long getHostSyncTimestamp() {
-    return hostSyncTimestamp;
+  private class HostData {
+    ArrayList<FrameData> framesMetaData;
+    long recordingStartTime;
+    long videoFps;
+    Long hostSyncTimestamp;
```
mwarton: Why is this a (boxed) Long when the others are (unboxed) long?

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/ServerHandler.java
```diff
@@ -1,125 +1,147 @@
...
-        outputWriter.flush();
-        Log.d(ContentValues.TAG, "Request sent to server. Waiting to receive timestamps..");
-        String message = inputReader.readLine();
-        serverStartTimestamp = Long.parseLong(message);
-
-        while (message != null) {
-          serverTimestamps.add(Long.valueOf(message));
-          Log.d(ContentValues.TAG, "Server TimeStamp:" + message);
-          message = inputReader.readLine();
+  private ArrayList<FrameData> parseTextObjects(ArrayList<Text> resultsOcr) {
```
mwarton: It would be better to make FrameData a public class and do this conversion in the class that handles the OCR.  That way you do not have any dependency from this class on the mlkit library, which makes it easier to change or add additional approaches if we ever need to do that.  It also makes it easier to test this class, as we don't need to be able to create an mlkit Text object.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -32,100 +30,76 @@
...
-        e.printStackTrace();
-        return null;
-      }
-    }
+  @RequiresApi(api = VERSION_CODES.N)
+  public CompletableFuture<ArrayList<Text>> doOcr() {
+    CompletableFuture<ArrayList<Text>> resultsOCR =
+        frameList.thenApply(
+            frames -> {
+              synchronized (frames) {
```
mwarton: This is the only code that synchronizes on frames, and it is only called once.  There is nothing to synchronize with, so you don't need the synchronize block.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -32,100 +30,76 @@
...
+                ArrayList<Text> ocrTexts = new ArrayList<>();
+                for (int i = 0; i < frames.size(); i++) {
+                  InputImage imageHolder = InputImage.fromBitmap(frames.get(i), 0);
+                  final int imageIndex = i;
+                  ocrTasks.add(
+                      recognizer
+                          .process(imageHolder)
+                          .addOnSuccessListener(
+                              visionText -> {
+                                ocrTexts.add(visionText);
```
mwarton: These callbacks may be executed out of order, you need to insert it in the correct location in the array rather than just appending.

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/VideoProcessor.java
```diff
@@ -32,100 +30,76 @@
...
+                              })
+                          .addOnFailureListener(
+                              e ->
+                                  Log.d(
+                                      ContentValues.TAG,
+                                      "Analyse failed with: " + e.getMessage())));
+                }
+                Task finalTask = Tasks.whenAllComplete(ocrTasks);
+                try {
+                  Tasks.await(finalTask);
```
mwarton: Add a comment that this is ok because this is on a background thread

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
```diff
@@ -330,10 +330,10 @@ class CameraFragment : Fragment() {
 
                     // Starts recording animation
                     overlay.post(animationTask)
+
+                    // Send signal to start key simulation in server text editor.
```
mwarton, line 335: Rather than telling the server what to do, it should just be sending a notification of what state it is in,  ie this should be "tell the server recording has started".  This makes it conceptually easier to use the app in a different context, eg starting a window and detecting the menu text.

Server/serverSlate.go
```diff
@@ -36,21 +59,49 @@ func main() {
...
-	message, _ := hostReader.ReadString('*') // using * as delim and hardcoded in app to append * at end of message
-	fmt.Println(message)
-	if message == "started capture"+"*" {
-		simulateKeyPress(KEY, KEY_PRESS_COUNT, timeStamps)
+	var hostAction action
+
+	for {
+		err := hostCommunicator.Decode(&hostAction)
+		fmt.Println(err)
+		if hostAction.Code == 1 {
```
mwarton: define constants for these command codes, eg ACTION_CAPTURE_STARTED and ACTION_CAPTURE_RESULTS

Server/serverSlate.go
```diff
@@ -36,21 +59,49 @@ func main() {
...
+func calculateLag(ocrData hostData, timestamps []int64) []int64 {
+	lagResults := make([]int64, KEY_PRESS_COUNT+1)
+	syncOffset := ocrData.HostSyncTimestamp - timestamps[0]
 
-	message, _ = hostReader.ReadString('*') // using * as delim and hardcoded in app to append * at end of message
-	fmt.Println(message)
-	if message == "send timestamps"+"*" {
-		sendTimeStamps(conn, timeStamps)
+	searchKey := ""
+	frameDuration := 1000 / ocrData.VideoFps
```
mwarton: maybe use a float, work in nanoseconds, or do the multiplicatoin by j before the division to avoid loss of precision here.

Server/serverSlate.go
```diff
@@ -67,18 +118,13 @@ func runOsCommand(cmd *exec.Cmd) int {
 	return 1
 }
 
-func simulateKeyPress(key string, keyPressCount int, timeStamps []string) {
-	timeStamps[0] = strconv.Itoa(int(time.Now().UnixNano() / 1000000))
+func simulateKeyPress(key string, keyPressCount int, timestamps []int64) {
+	timestamps[0] = time.Now().UnixNano() / 1000000
```
mwarton: Can you store this in a different variable and return it as a second value?  This would clean up the iteration code in the calculateLag function too.

Server/serverSlate.go
```diff
@@ -67,18 +118,13 @@ func runOsCommand(cmd *exec.Cmd) int {
 	return 1
 }
 
-func simulateKeyPress(key string, keyPressCount int, timeStamps []string) {
-	timeStamps[0] = strconv.Itoa(int(time.Now().UnixNano() / 1000000))
+func simulateKeyPress(key string, keyPressCount int, timestamps []int64) {
```
mwarton: please return timestamps rather than using an "out" parameter

Camera2SlowMotion/app/src/main/java/com/example/android/camera2/slowmo/fragments/CameraFragment.kt
aquatiko: Yeah that is implemented using ACTION codes trasmitted to server.

Server/serverSlate.go
```diff
@@ -14,7 +14,9 @@ import (
 
 const KEY = "m"
```
mwarton: minor: use const (
KEY = "m"
KEY_PRESS_COUNT = 410
...
)

Server/serverSlate.go
```diff
@@ -23,24 +23,24 @@ type action struct {
...
 
-type point struct {
+type Point struct {
 	X int `json:"x"`
 	Y int `json:"y"`
 }
 
-type frameData struct {
-	CornerPoints []point `json:"cornerPoints"`
+type FrameData struct {
```
mwarton: You shouldn't need to make these public just for testing.  go unit tests have access to private stuff in the package they live in.

Server/serverSlate_test.go
```diff
@@ -1,7 +1,29 @@
...
+	"fmt"
+	"io/ioutil"
+	"testing"
+)
 
-func testServerSlate(t *testing.T) {
-	// TODO: Add serverSlate tests
+func TestCalculateLag(t *testing.T) {
+	file, err := ioutil.ReadFile("ocr-data.json")
+	fmt.Println(err)
```
mwarton: Check err properly
if err != nil {
 t.Fatal("Unable to read ocr-data.json: %v", err)
}

Server/serverSlate_test.go
```diff
@@ -1,7 +1,29 @@
...
+	var ocrData HostData
+	json.Unmarshal([]byte(file), &ocrData)
```
mwarton: check the error

Server/serverSlate_test.go
```diff
@@ -1,7 +1,29 @@
...
+
+	testStartTimestamp := int64(1597670880210)
```
mwarton: If you are reading some data from a file, you probably should read all of it.

Make a higher level structure, eg:

type LagTestData struct {
  startTimestamp int64
  keyTimestamps []int64
  ocrData HostData
  cameraStartDelay int64
  expectedLag []int64
}

mwarton (approved):

Server/serverSlate_test.go
aquatiko: added to the same json file.

Server/serverSlate_test.go
aquatiko: done

Server/serverSlate_test.go
aquatiko: done

Server/serverSlate.go
aquatiko: Didn't knew about this, fixed!

Server/serverSlate.go
aquatiko: changed!

mwarton (approved):

